PR #224, opened 2013-09-20
Reset historical CPU usage data if delta is negative for individual cores.

filipeaneto:


chrisspen:
I'm not sure how the code change accomplishes what you're describing. The change to support Gnome 3.18 has been merged in another pull request.

Issue #207, opened 2013-04-19
Issues with gvfs interaction


camelreef:
Hello,

Ubuntu 12.10 and 13.04 with the Gnome 3 PPA, Gnome 3.6 and 3.8
- the system does not go fully into suspend and comes back to the shell, sometimes into a broken state if there is a gvfs mount
- shell fully frozen while doing a file transfer using the MTP gvfs backend
- shell freeze if mounted iPhone is removed
- regular shell/UI freezes

All of this disappears if the extension is disabled

It's a shame, I love the extension otherwise!

cryptoluks:
I can confirm this issue.
Arch Linux x86_64 3.18.7-2-ck
GNOME Shell 3.14.3

daptordarattler:
I can confirm this issue with Ubuntu 17.10 and latest gnome-shell and gnome-system-monitor. Helped my find a hidden gvfs I didn’t kno about.

chrisspen:
Can anyone confirm this problem exists on Ubuntu 18.04 or later? I'm experiencing no issues.

jjcf89:
Running 18.04 and my gnome-shell freezes up fairly regularly.  Though I can't guarantee its still this issue like it was in the past.

chrisspen:
And the problem goes away if you disable this extension?

jjcf89:
I don't have an easily reproducible way to trigger this to test for sure.

return42:
@chrisspen, @jjcf89 .. can we close this thread and continue discussion in #226  (started 2015)

@camelreef if it is OK for you, can you please close this issue (started 2013)?

return42:
@chrisspen @jjcf89 @camelreef  .. forget my last post .. I was a bit confused .. sorry for the noise.

return42:
I can confirm, that the extension works with ubu18-04. / lets close this old issue.

jjcf89:
> I can confirm, that the extension works with ubu18-04. / lets close this old issue.

The fact that the extension works in 18.04 is hardly proof that this bug has been fixed as I don't think the bug resolves around whether the plugin works or not.  But whether it hangs the system based on gvfs interactions.

Issue #496 does seem to resemble the issue.

return42:
[...]

PR #257, opened 2014-11-15
Avoid freezes on SFTP/MTP mounts

fxthomas:
Workaround for #207 when copying files from/to SFTP or MTP mounts.

Is there a reason why we can't use the [is_native()](https://developer.gnome.org/glibmm/stable/classGio_1_1File.html#aaa3b910bc9f240d4ebd0efc70979f163) method there? It seems to me that it would handle most edge cases properly, and both SFTP and MTP would have been handled by this.


darkxst:
is_native seems like a good idea, can someone test and check that it works as expected?

chrisspen:
Rejecting due to @kartikmohta's comment, and since this appears to have been added in a separate ticket.

PR #265, opened 2015-01-27
Better support when NetworkManager is not present

fthiery:
The extension will now look for static interfaces every 3 seconds until it finds one that is up. 

My system doesn't use NetworkManager (but systemd-networkd) and my network is probably slow to get up, so when i enter gnome, the link isn't up yet and i get no network stats unless i restart gnome itself.

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -1433,8 +1433,6 @@ const Net = new Lang.Class({
         this.ifs = [];
         this.client = NMClient.Client.new();
         this.update_iface_list_nm();
-        this.update_iface_list_static();
-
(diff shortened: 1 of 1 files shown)


darkxst:
log messages should work fine from the callback, are you getting any log messages at all?

I am not 100% sure that the return values will make it back through the closures in JS, but test it and if it does its fine.

Also Network Manager should know about all connections these days, even static connections setup with ifup or networkd. Although I havent checked if bandwith data is available in these situations. Disabling network manager completly is likely a bad idea since its a hard dep of GNOME and certain parts will expect it to be running even if its not managing any interfaces.

darkxst:
btw If the timeouts aren't getting cancelled automatically then just use Mainloop.source_remove(id)

chrisspen:
Closing this, due to no activity in years. Please feel free to resubmit if you can resolve the conflicts.

PR #281, opened 2015-08-18
Add FreeBSD support

lantw44:
This pull request changes following things:
1. Detect currently running operating system at runtime instead of assuming we always use Linux.
2. Make NetworkManager optional on non-Linux systems.
3. Call glibtop_init if it is available in `enable()` to prevent odd crashes.
4. Make it possible to write system-dependent code in C.
5. Port Freq and Net element to FreeBSD.

Current status on FreeBSD:
- Working: CPU, Memory, Swap, Net, Disk, Freq
- Broken: Thermal, Fan
- Not tested: Battery

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,7 @@
 *~
 tmp
+*.gir
 *.pyc
+*.so
 *.swp
+*.typelib
--- a/system-monitor@paradoxxx.zero.gmail.com/Makefile
+++ b/system-monitor@paradoxxx.zero.gmail.com/Makefile
@@ -1,0 +1,43 @@
+.PHONY: all clean distclean
+
+# Commands
+INTROSPECTION_SCANNER = g-ir-scanner
+INTROSPECTION_COMPILER = g-ir-compiler
+
+# Arguments
+SMSYSDEPS_CFLAGS = -fPIC -shared -Wall -O2
+
+# Sources
+CFILES = sysdeps/system-monitor-sysdeps.c
+HFILES = sysdeps/system-monitor-sysdeps.h
+
+# Compiled shared object
+SOLIB  = system-monitor-sysdeps
+SOFILE = sysdeps/lib$(SOLIB).so
+
+# Compiled gir and typelib file
+NAMESPACE = SystemMonitorSysdeps
+NSVERSION = 0.1
+SMPREFIX = system_monitor_sysdeps
+GIRFILE = sysdeps/$(NAMESPACE)-$(NSVERSION).gir
+TYPELIBFILE = $(GIRFILE:.gir=.typelib)
+
+all: $(TYPELIBFILE)
+distclean: clean
+
+$(SOFILE): $(CFILES) $(HFILES)
+	$(CC) $(SMSYSDEPS_CFLAGS) $(CFLAGS) $(CFILES) -o $(SOFILE) $(LDFLAGS)
+$(GIRFILE): $(SOFILE)
+	LD_LIBRARY_PATH=./sysdeps:${LD_LIBRARY_PATH} $(INTROSPECTION_SCANNER) \
+		$(CFILES) $(HFILES) \
+		-Lsysdeps \
+		--library=$(SOLIB) \
+		--namespace=$(NAMESPACE) \
+		--nsversion=$(NSVERSION) \
+		--symbol-prefix=$(SMPREFIX) \
+		--output=$(GIRFILE)
+$(TYPELIBFILE): $(GIRFILE)
+	$(INTROSPECTION_COMPILER) $(GIRFILE) -o $(TYPELIBFILE)
+
+clean:
+	rm -f $(SOFILE) $(GIRFILE) $(TYPELIBFILE)
--- a/system-monitor@paradoxxx.zero.gmail.com/Makefile
+++ b/system-monitor@paradoxxx.zero.gmail.com/Makefile
@@ -3,9 +3,14 @@
 # Commands
 INTROSPECTION_SCANNER = g-ir-scanner
 INTROSPECTION_COMPILER = g-ir-compiler
+PKG_CONFIG = pkg-config
 
 # Arguments
-SMSYSDEPS_CFLAGS = -fPIC -shared -Wall -O2
+SMSYSDEPS_PKGS = glib-2.0
+SMSYSDEPS_PKGS_CFLAGS = `$(PKG_CONFIG) --cflags $(SMSYSDEPS_PKGS)`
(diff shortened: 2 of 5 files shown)


lantw44:
Rebased for GNOME 3.18 support.

darkxst:
@lantw44 hey thanks for the patches, I will try and review them later in the week

lantw44:
My patches have stayed here for several months. Can them be reviewed?

darkxst:
Sorry for the delay!

I would really rather not ship a C wrapper program with the extension. Is there absolutely no other way to get this information?

lantw44:
Unfortunately, gjs doesn't provide binding to FreeBSD sysctl interface, and I can't find a way to directly call sysctl in gjs. Its function prototype is
```
int sysctl(const int *name, u_int namelen,
           void *oldp, size_t *oldlenp,
           const void *newp, size_t newlen);
int sysctlnametomib(
           const char *name, int *mibp, size_t *sizep);
```
The type of `oldp` and `newp` depends on the name array, and `mibp` is an array of size `*sizep`. Both are hard to bind in gjs, and I spent two days trying to generate and write a gir file without success.

I think it may be possible to get some information via command line tools such as `sysctl` and `ifconfig`, but I want to avoid fork and exec because this creates noise for users who use `execsnoop` or `dtrace` to debug their systems.

`sysctl` command line tool is not as powerful as `sysctl` system call, and FreeBSD doesn't provide pseudo filesystems like `/proc` and `/sys` on Linux. `getifaddrs` function is implemented using `sysctl` system call, but `sysctl` command line tool is unable to do the required `sysctl` system call. It may be possible to get required information by parsing output of `ifconfig`, but I want to avoid parsing complex things if possible.

p91paul:
I believe that shipping C code is a bad idea; however, if no js solution is available, the C helper code could be wrapped in a library, possibly hosted in a separated repository; the js support could be embedded in this extension, which would then add the external library to its FreeBSD dependencies.

lantw44:
The C helper code in my patch is already a library specifically made for this extension. It should be compiled as `sysdeps/libsystem-monitor-sysdeps.so` in the extension directory. As the library is not useful for other applications, I didn't want to install this library to the system directory. I also didn't want to ask users to provide the path to the library, so I put it in the extension directory. Do you mean I should move the library file out of the extension direcotory, and add code to `extension.js` to find it?

p91paul:
I see you wrote your own Makefile. I suggest the following:
* Switch to autoconf. It'll provide automatically a --prefix option to ./configure for your users to choose the install directory
* Once the above is done, don't worry about the install directory. The standard is to install system-wide, so follow the standard; your users will be able to change that if they prefer to install into some other directory; there is plenty of single purpose libraries, don't be ashamed to put yours in.
* If everything is installed in standard directories, you won't need to add to gir search path and library path manually in the extension.
* A bit ignorant on this point: does gir have a way to specify additional search paths through modifications of configuration files or env variables (I mean like LD_LIBRARY_PATH acts on libraries)? If the answer is yes, let your users take care of that (obviously documenting the needed steps in a README). If it is no, add a preference in the extension preference window (possibly displayed on FreeBSD only) to specify the user-specified --prefix;

This is a lot of work, but IMHO is the cleaner approach; of course, I'm not the maintainer, so this is just friendly advice :)

darkxst:
I agree with Paul on this.

* Put the script in its own repo, or at the very least in its own folder on the main repo, this is not something we could ever ship on the extensions.g.o site for example.

* autoconf/automake are much nicer than handcrafted Makefiles, make things easier for packagers as well. Though if you have never used them before, the learning curve might make it overkill for such a simple project.
* Install by default in system dirs, but ~/.local/lib is a good fallback for people that want it installed on their user account only. that *should* be picked up automatically.
* I have no issue including the JS Side of things into the main branch, so long as its properly fenced from affecting other distro's (I think it mostly was, but haven't done a full review of the code just yet)

lantw44:
The C wrapper I made is a private library, a library with unstable interfaces that can only be used by one application. This kind of library is usually installed into its private directory, like what `gnome-shell `does for its `libgnome-shell.so`. (installed in `/usr/local/lib/gnome-shell`, not the system-wide `/usr/local/lib`) I didn't make a general sysctl library because it will be much complex, requiring interpreting bytes as larger integers or kernel structs in JS.

I didn't use autotools because there is only single C source file and the build process can't be more complex for non-Linux systems supported by gnome-shell (currently FreeBSD and OpenBSD). Using autotools also introduces more dependencies for users who download the extension from extensions.gnome.org or github. With a Makefile, only `pkgconf` has to be installed. With autotools, users have to install `autoconf`, `automake`, `libtool`, `pkgconf`, `gmake` because a `configure` script is not included in the git repository and `Makefile.introspection` is incompatible with BSD make.

I agree that autotools is much easier for packagers, and writing the installation prefix to a .js file with information from autotools is an acceptable solution although users have to remember not to move the directory after `gmake install`. I will get this extension packaged in FreeBSD after patches are accepted, so I hope most users install the binary package instead of downloading it from extensions.gnome.org or github.

Before I start adding autotools support and moving sysdeps out of the extension directory, I want to make sure the following source and installed trees are acceptable:

Source:
```
autogen.sh
configure.ac
Makefile.am
sysdeps/system-monitor-sysdeps.c
sysdeps/system-monitor-sysdeps.h
system-monitor@paradoxxx.zero.gmail.com/
<other files>
```

Installed:
```
lib/gnome-shell-system-monitor-applet/libsystem-monitor-sysdeps.so
lib/gnome-shell-system-monitor-applet/SystemMonitorSysdeps-0.1.typelib
share/gnome-shell/extensions/system-monitor@paradoxxx.zero.gmail.com/
```

typelib search path can be modified by `GI_TYPELIB_PATH` environment variable, but I don't think it is a good solution because the variable is useless for applications launched by gnome-shell.

darkxst:
I would prefer that the C wrapper lives in a subfolder that is clearly marked with Freebsd or bsd in the name just to avoid any confusion. We don't want people downloading the git repo and then thinking they need to build sysdeps on linux etc! 

To be honest I am not a huge fan of the extensions being packaged in distro's, but it does tend to happen. So it would probably be ok to have a autoconf target that just adds a system install of the extension, in the root of the git repo. Your proposed installed paths look fine to me

As far as extensions.g.o goes though, that just has a snapshot of the system-monitor@paradoxxx.zero.gmail.com folder. I doubt they would even allow the wrapper to be shipped in that snapshot (and I probably wouldn't want to either)

lantw44:
If C wrappers are not allowed on extensions.gnome.org, FreeBSD users are not be able to install this extension with extensions.gnome.org, so I think getting it packaged is needed to make installation easy.

chrisspen:
Closing this, due to no activity in years. Please feel free to resubmit if you can resolve the conflicts.

PR #283, opened 2015-09-22
Fix manual install in 3.16: needs to enable extension with gnome-shell-extension-tool

brunetton:
I had a problem with Gnome shell 3.16 and automatic install via extensions.gnome.org : extensions doesn't install and nothing in logs. I didn't searched a lot around and tried manual install.
After manual install procedure, extension wasn't enabled; I needed to enable it (via *https://extensions.gnome.org/local/* or *gnome-shell-extension-tool*), so adding it in manual install procedure seems to be a good idea, don't you think ?


pwaller:
Works for me, thanks! (Was puzzled that I couldn't get it to work)

darkxst:
yes, thanks, that is the right procedure. Merged

PR #285, opened 2015-09-23
support for GNOME 3.18

ousia:


lantw44:
Does this really work? St.Table was removed in 3.17.2:
https://git.gnome.org/browse/gnome-shell/commit/?id=86e0404

p91paul:
Damn, St.Table code was introduced in this extension as of 3.12 (https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/pull/251) because previous implementation broke. I see Gnome developers still pay some attention to make sure they break APIs every few releases, even if this is the 10th Gnome 3 release.

samueljun:
Does anyone know what implementation should be used instead?

p91paul:
The commit mentions this bug report: https://bugzilla.gnome.org/show_bug.cgi?id=703833, where they talk about removing StTable in favour of ClutterTableLayout. Attached to that bug there are patches where references to St.Table are replaced with St.Widget with `{ layout_manager: new Clutter.TableLayout() }`. Here is an example: https://bugzilla.gnome.org/attachment.cgi?id=250036&action=diff

p91paul:
Good news is that, according to the bug report I linked, work on removing StTable dates back to 2013, so hopefully switching to ClutterTableLayout won't break the extension for older gs versions because it already existed back then.

chrisspen:
Resolved by #288.

PR #296, opened 2015-11-28
Updated italian translation

TopoDiFogna:
I've updated the translation and added missing ones


po/it/system-monitor.po
```diff
@@ -1,7 +1,7 @@
 # Traduzione italiana per il pacchetto system-monitor.
 # Copyright (C) 2012 THE system-monitor'S COPYRIGHT HOLDER
 # This file is distributed under the same license as the system-monitor package.
-# Federico Piazza fede.piazza92@gmail.com, 2012.
```
p91paul: Differently from the Last-Translator, I would not remove this attribution. Just add yours!

po/it/system-monitor.po
```diff
@@ -207,7 +207,7 @@ msgid "other"
 msgstr "altro"
 
 msgid "nice"
-msgstr "nice"
```
p91paul: AFAIK, "nice" is a technical term and isn't translated in Italian. I'm Italian, but I never heard of "particolare" as a translation for "nice"; can you point to a reference for that?

TopoDiFogna: Nice is used as detailed in this case i think, particolare or particolareggiato should do the job

po/it/system-monitor.po
TopoDiFogna: Added one more commit to restore previous translator as requested

po/it/system-monitor.po
p91paul: Check [this](http://askubuntu.com/questions/399357/what-does-the-nice-value-mean-in-cpu-utilization-statistics): nice is the time spent on niced (i.e. high priority) tasks. I believe it should not be translated!

Issue #259, opened 2014-11-25
Spamming system logs due to non-existent hwmon files


mgalgs:
I get this in my system log every 5 minutes:

```
Nov 25 11:02:14 mitchelh-linux gnome-session[814]: Gjs-Message: JS LOG: error reading: /sys/class/hwmon/hwmon1/fan1_input
```

This makes it very difficult to read my system logs.

I had a look at `check_sensors` but it looks like we're already bailing out if the files don't exist...   So I'm not sure where the problem is.

Hecke:
me too...

though it is all 2-3 secs and freezes my system for a short time :-(

Jan 27 09:06:39 arenal gnome-session[1440]: Gjs-Message: JS LOG: error reading: /sys/class/hwmon/hwmon1/temp1_input

najamelan:
I have the same problem

darkxst:
this should been fixed by
https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/commit/d04c1366c53069b2e59b875a51878cf3a8d78b8a

tommybutler:
I'm running Debian Jessie, and compiling from source isn't really feasible.  How, please tell me, how do I turn this off?  What process needs to be killed or whatever?  I can't just touch an empty file with that name into the sys filesystem to make the complaint go away... Please, seriously please tell me how to make it stop.  Bandaids and duct tape fixes are perfectly acceptable for me.

p91paul:
@tommybutler There is no "compiling from source" involved, javascript extensions are interpreted without needing to be compiled and are installed to your home folder, so you don't need administrative permissions. You just have to click on the "Download ZIP" button in the homepage of this project, unzip the file, and copy the extracted system-monitor@paradoxxx.zero.gmail.com folder in /home/<your user>/.local/share/gnome-shell/extensions/, replacing files if you have the extensions.gnome.org version installed. Please remember that .local is an hidden folder, so if you are using your file manager you'll have to set hidden files and folders to be visible to see it.

Anyway, detailed manual install instructions (slightly different from the ones I am providing here, mainly because they require installing git) are also provided on the project homepage.

tommybutler:
Thank you @p91paul as I was unaware these extensions were javascript code.  I appreciate the tip, it got me fixed right up and now my syslog isn't bloating out anymore.  Thanks again!

tommybutler:
[...]

PR #299, opened 2015-12-29
Gnome 3.8 Branch for EL7 - fix for noisy logs when hwmon devices are not available

salderma:
PR to fix issue #259 for Gnome 3.8 and provide better support to EL7 users.  I'd suggest taking this PR off as a new branch similar to what you've done with gnome-3.0 and gnome-3.2, although I'm not sure how you do that exactly.

This branch was picked up at the most recent commit (cc86a28) which works for Gnome 3.8.


darkxst:
thanks! done

PR #302, opened 2016-04-02
Enable Gnome 3.20

marxin:
I've just tested the applet on openSUSE Tumbleweed with gnome 3.20.


Majkl578:
Works fine on Debian as well.

HarlemSquirrel:
Works great on Antergos so far!

PR #312, opened 2016-08-08
fixing frequency graph display, and improving efficiency of the frequency check

slowriot:


chrisspen:
I'm a little confused. What does this change do exactly? You're setting an array index which is almost immediately overwritten.

chrisspen:
Feel free to re-open this if I'm missing something.

Issue #279, opened 2015-07-22
Allow memory stats to be displayed in GB


mrthehud:
Instead of seeing that I have 15494 MiB of ram, I'd like to be able to see "16.5 GiB" by selecting either "GiB" or "Ideal" (Uses appropriate unit to keep value between 0 and 1023) from a "Unit" dropdown on the memory tab in the preferences panel.

Or is this possible already using dconf or some other tool?

aside: I think Ubuntu switched to using SI data units for data, eg MB (10^6) instead of MiB (2^20) a few years ago, so is it (or can it be) possible to tell the whole applet to use either SI or Binary units?

marxin:
I guess this issue can be closed as the pull request was merged.

PR #313, opened 2016-08-11
Use GiB instead of MiB when we have more than 4GiB of RAM (or swap)

franglais125:
Concerning issue: https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/issues/279

This commit uses GiB as the unit for RAM and swap whenever the size is larger than 4GiB.

The threshold can obviously be changed (lines 1400 and 1641).
4GiB is maybe a low value, but when using something like ~16GiB of RAM, it is clearly more user friendly to have GiB as the unit.

Also, by default I use two decimal places. That could be adjusted as well.


chrisspen:
Thanks for the commit.

franglais125:
@chrisspen Thanks for all the work, you are a legend!

PR #320, opened 2016-12-03
Fix preference display order

alexheretic:
Common preferences of Cpu, Memory, etc appear in different orders, which I find a little annoying.

This patch should ensures that the first 2 horizontal layout boxes (which are most common) are ordered alphabetically by label.
Ie in all preference sections we'll have 'Display', then 'Show In Menu', then 'Show Text'.

Also added some simple readability improvements & style updates I believe are in accordance with the gnome style guide but I can change that if they don't match the project's style preferences.


alexheretic:
I'm going to move this change to a branch and continue changes in my master

PR #322, opened 2016-12-09
Major code refactor: split extension.js

alexheretic:
I started looking into fixing the menu styles as I found it a little cramped, but found the code to be a little difficult to read and very easy to break. I think the code can benefit greatly from being split into separate files for each of the logical areas, Cpu, Memory etc. Since these are already represented with classes this is natural, and despite the poor GJS extension import system, works well.

Since almost everything was shoved in the extension.js we have a lot of inconsistent and non-ideal Javascript code. I trimmed dead code, fixed obvious bugs & introduced more es6 style Javascript style when the wins were fairly easy. Since this large bunch of code has been split out, the PR is fairly major. However, the code should function as before (although I'm only testing on the latest version of gnome).

I believe this change makes the code easier to work with in isolation (say just working on the Cpu logic) in confidence that changes haven't broken other areas. Thus it should improve the quality of future additions to the code base.

~~With the js code-split complete I did make my style changes to the menu. I switched to network up & down icons, rather than text (similarly to the current readme screenshot), and fixed the unit alignment to be consistent. Also brightened the main labels, and the actual values against the now brighter background.~~
~~Finally the graph colours are tweaked so the initial colour is discernible against the background.~~
~~Please see the before and after (running on gnome 3.22.2, arch)~~
Remove for subsequent PR


alexheretic:
I've merged the latest repo merges into my branch, so the split should be up to date again. You can just tell me if you're aren't interested in the split, but I think it's an improvement.

franglais125:
@alexheretic actually it's better if you _rebase_ your branch, instead of merging the main one into yours.

I made the same mistake a few times on other projects. It is harder for people to see what your changes were: github shows the changes made by all the commits you merged (see the link for "Files changed" at the top).

Assuming you don't know how to do it, I had to research that myself at first:

1) Add upstream
`git remote add upstream https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet`
2) Create a new branch, just in case there is a problem
`git checkout -b rebased_branch`
3) Remove the last commit (i.e. the merge)
`git reset --hard HEAD^`
4) Rebase!
`git rebase upstream/master`
5) Done. Push your new branch to your repo!

alexheretic:
Hi, the files changed looks accurate to me its a big code split that affects most of the source. The merge has worked normally (ie you don't see any of the po/* changes, for example po/it/system-monitor.po was merged)

The code split has effected extension.js, and previously pretty much all the js logic was there. Automatic merge/rebase can't handle the move automatically with changes made to the previous layout. I simply manually merged the extension.js changes into the relevant split files in model/. Since the merged changes were so small this was easy.

To summarise they show up as changes as they are in different files now, refactored to me more intendant modules. I hope that makes sense!

franglais125:
Indeed, I fully understand what you are saying. I went through the pain of rebasing many times, and I still think it's worth it. I tried rebasing your branch to see what would happen, and sure enough there are conflicts. I do know it's not straightforward!

If you look at your current master [1], you see that the effect of the last merge is that your commits are applied_before_ the merges that already took place. From the maintainer's perspective, this looks non-linear, and it is later on much more difficult to track down what really changed.

Anyway, just my point of view... I'm not the one merging anything here :).

And thanks for taking the time to produce that huge refactoring!

[1] https://github.com/alexheretic/gnome-shell-system-monitor-applet/commits/master

alexheretic:
Right fair enough, I've squashed my changes into a single commit on top of current master. Thanks for the help.

darkxst:
@alexheretic While splitting makes sense and is probably worth doing, having it as a single commit is impossible to review!

I would suggest the first commit should just have the mechanical split and subsequent commits with your other changes.

alexheretic:
I can understand that, I've removed the minor style updates. Now this PR is just the major Javascript code style update. I hope it helps you merge it. 

One thing worth mentioning is merging after this split should become easier, with conflicts confined to the files they affect rather than all affecting extension.js

darkxst:
@alexheretic What I would suggest is to make your first commit just the code split. i.e. literally just moving the existing code into the different files without any code changes at all (even though it obviously won't work at this point). Once you have that as a base apply the code changes that you have made as separate commits over that.

That will provide a nice history that clearly shows the changes you have made, and will make reviewing your patchset match easier!

alexheretic:
@darkxst I have it as two commits now. Should be somewhat easier to review the per-file diffs now, [with whitespace disabled](https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/pull/322/commits/51c4930486b26e1fe237eb2eafd97d60220eaea1?w=1). However, its still a big refactor and I think its just going to take some effect to review. If you guys don't think its worth it, that's your call. I don't think I can do much more with squashes and shuffling around branches to make the split easier to digest (short of re-coding the entire thing). I am happy to explain the reasoning behind any of the non-structural changes though.

alexheretic:
Hey perhaps this one has proved too hard to merge in. _Are_ you guys interested in having the code split this way? If so, maybe we could stagger it so we have a PR for each class extracted out from extension.js.

Ie we could start we everything the same, but just the `Mem` class moved to model/memory.js and imported to extension.js.
```js
// model/memory.js
(exports => {
    exports.constructor = new Lang.Class({
         Name: 'SystemMonitor.Mem',
         // ...
    });
})(this);
```
```js
// extension.js
const local = imports.misc.extensionUtils.getCurrentExtension().imports;
const Mem = local.model.memory.constructor;
```
What do you think?

chrisspen:
@alexheretic As the only active maintainer, yes, as it stands, it's unlikely to be merged anytime soon. I don't have the time nor the intimate familiarity with the code to handle the scope of these changes. And as the code isn't being actively developed anymore, aside from minor bugs fixes, a large refactoring doesn't seem justified.

I hope these comments don't dampen your enthusiasm. You've clearly put a lot of work into it, and I appreciate that, but this project is in maintenance mode, and unfortunately, no one has time to give your considerable changes the proper testing they require.

alexheretic:
Ok @chrisspen thanks for the reply.

PR #336, opened 2017-01-31
Update Brazilian Portuguese translation

josephgbr:


chrisspen:
Thanks for the commit.

PR #338, opened 2017-02-10
Disk: fix I/O speed unit labels' size

develop7:
Fixes `MiB/s` label size in applet tooltip.

Env: GNOME 3.22

---
This change is [Reviewable](https://reviewable.io/reviews/paradoxxxzero/gnome-shell-system-monitor-applet/338)

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -1331,10 +1331,10 @@ const Disk = new Lang.Class({
     },
     create_menu_items: function() {
         return [new St.Label({ style_class: Style.get("sm-value")}),
-                new St.Label({ text:_('MiB/s'), style_class: Style.get("sm-label-left")}),
+                new St.Label({ text:_('MiB/s'), style_class: Style.get("sm-label")}),
                 new St.Label({ text:_('R'), style_class: Style.get("sm-label")}),
                 new St.Label({ style_class: Style.get("sm-value")}),
-                new St.Label({ text:_('MiB/s'), style_class: Style.get("sm-label-left")}),
+                new St.Label({ text:_('MiB/s'), style_class: Style.get("sm-label")}),
                 new St.Label({ text:_('W'), style_class: Style.get("sm-label")})];
     },
 });


franglais125:
Maybe I don't see the same thing as you do, due to the theme, fonts, etc.
But with your patch, there is no separation between the unit (MiB/s) and the description (R or W). The original looks better IMO... Is it different for you?

**Original**
screenshot from 2017-02-10 15-04-36


**With your patch**
screenshot from 2017-02-10 15-05-15

develop7:
Before patch:
screenshot from 2017-02-10 23-27-00

After patch:
screenshot from 2017-02-10 23-24-44

franglais125:
Wow! Huge difference, maybe it's due to the locale?

develop7:
Sorry I didn't posted screenshots before — got little time.

Anyway, consider "Net" section:

```js
    create_menu_items: function() {
        return [new St.Label({ style_class: Style.get("sm-value")}),
                new St.Label({ text:_('KiB/s'), style_class: Style.get("sm-label")}),
                new St.Label({ text:_('Down'), style_class: Style.get("sm-label")}),
                new St.Label({ style_class: Style.get("sm-value")}),
                new St.Label({ text:_('KiB/s'), style_class: Style.get("sm-label")}),
                new St.Label({ text:_('Up'), style_class: Style.get("sm-label")})];
    }
```

Couldn't help but notice the layout is identical to "Disk", so I figured replacing `sm-label-left` to `sm-label` seems acceptable.

>  maybe it's due to the locale?

It shouldn't be, more like a bug: both `sm-label` and `sm-label-left` got identical explicit `font-size`s.

develop7:
@paradoxxxzero what'd be the verdict?

paradoxxxzero:
After a quick look, Net and Disk units should definitely have the same classes. As for why `sm-label-left` renders weirdly I don't know, do you use the default theme?

develop7:
Yes, both shell and GTK+ themes are "Adwaita"

develop7:
> Net and Disk units should definitely have the same classes

And now they do! :)

paradoxxxzero:
Let's go with that and maybe add a padding/margin one way or another to add space between labels

franglais125:
@chrisspen sorry for the push lately. Just a ping about this (small) PR!

waldyrious:
> with your patch, there is no separation between the unit (MiB/s) and the description (R or W). The original looks better IMO... Is it different for you?

FWIW I already see it that way (without the space between `MiB/s` and `R`/`W`):

screenshot from 2018-02-28 18-06-14

I was actually going to submit an issue or PR until I found this :)

franglais125:
We sort of lost track of this PR, but from what I can see this issue is now fixed. Or at least the patch worked its way into master, as the correct styles are being used... I'll close this for now, please reopen if the issue is still not fixed.

Cheers, and thanks!

PR #343, opened 2017-02-21
Ignore interfaces like *br* in Net.update_iface_list also

thomas-wach:
When building current iface-list within class Net _init from /proc/net/dev, bridge-interfaces are excluded to avoid double count of traffic.
But this isn't done within class function update_iface_list, where the class gets its interfaces via NMClient.Client.


chrisspen:
Thank you for the contribution, but I'm closing due to conflicts.

PR #346, opened 2017-03-25
Support Gnome 3.24

MichaelAquilina:
Add support for Gnome 3.24


MichaelAquilina:
Was under the impression this was working, but actually this fails on restart after an upgrade. Sorry about that

PR #347, opened 2017-03-26
Fix runtime error in GNOME 3.24

### LINKED ISSUE #341: extension.js tweak for  gnome-shell 3.24
Ubuntu Zesty use gnome-shell 3.23.90 (3.24) version now
but that extension is not yet compatible;

GNOME Shell 3.24 now uses mozjs38 (SpiderMonkey 38) which has stricter syntax rules.
https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey

 please do asap
Thanks

develop7:
Declare GTop and NetworkManager outside of try-catch scope

Fixes #341


chrisspen:
Thanks for the patch.

develop7:
it's @franglais125, I've only tested it :)

PR #352, opened 2017-04-06
Add CPU frequency max, min and average graph

slowriot:
Modern CPUs run each core at a different frequency; a single frequency graph doesn't make sense for many-core systems.  This PR adds a max, min and average plot for frequencies.


chrisspen:
Thank you for contributing, but I'm closing due to conflicts and lack of activity.

PR #354, opened 2017-04-06
Update translation files

franglais125:
I also added `ar/` and `ca/`, which were not included originally for some reason.


franglais125:
@chrisspen I added the compilation of translations to the `_build` part, so it gets executed when installing or when building the zip file for upload to e.g.o.

Issue #360, opened 2017-04-16
Decimal values should take care of the locale


DevDef:
All displayed decimal values should take care of the locale and have the correct decimal and digit group separators.
Now they are using default C locale, but lots of countries use the comma as the decimal separator and a thin space as the digit group separators (see: https://en.wikipedia.org/wiki/Decimal_mark).

DevDef:
Replacing all the calls to the Number.toString() method by Number.toLocaleString() is supposed to do the job.

PR #364, opened 2017-04-23
French translation update

DevDef:
Please pull in this French translation update

### DIFF
--- a/po/fr/system-monitor.po
+++ b/po/fr/system-monitor.po
@@ -254,9 +254,13 @@ msgstr "Kio/s"
 msgid "MiB/s"
 msgstr "Mio/s"
 
-#: extension.js:1419 extension.js:1635
+#: extension.js:1419 extension.js:1482 extension.js:1635
 msgid "MiB"
 msgstr "Mio"
+
+#: extension.js:1484
+msgid "GiB"
+msgstr "Gio"
 
 #: extension.js:184
 msgid "kB"
--- a/po/fr/system-monitor.po
+++ b/po/fr/system-monitor.po
@@ -1,21 +1,23 @@
 # French translation.
-# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
-# This file is distributed under the same license as the PACKAGE package.
+# Copyright (C) 2011 Florian Mounier aka paradoxxxzero
+# This file is distributed under the same license as the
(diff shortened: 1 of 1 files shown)


chrisspen:
Thanks for the contribution. However, are you sure about the translation for "nice"? `Nice` in this context is the name of a program and a proper noun, not the adjective "nice".

franglais125:
I don't feel strongly about it. While `nice` is the kernel utility, "politesse" refers to the concept [1], so I think it's ok to translate the string.

@DevDef the only thing that I'm not sure about is modifying "R" and "W" for "en lecture" and "en écriture". The reason is that it clutters a bit the menu. Maybe a shorter version? Like "lecture" and "écriture"? Also, thanks for the PR :)

[1] http://www.man-linux-magique.net/man2/nice.html

DevDef:
> I don't feel strongly about it. While nice is the kernel utility, "politesse" refers to the concept [1], so I think it's ok to translate the string.

I found that “politesse” was a pretty elegant translation (already seen elsewhere). ;-)

> @DevDef the only thing that I'm not sure about is modifying "R" and "W" for "en lecture" and "en écriture". The reason is that it clutters a bit the menu. Maybe a shorter version? Like "lecture" and "écriture"?

@franglais125: With such an alias, I guess you are French or, at least, French is your native language. So, I’m sure that you will find really ugly to have something like “12 Mio/s lecture”.
I didn’t find something shorter than “en lecture” and “en écriture”. But, you know, there’s already “descendant” and “montant” on the line above, so it’s almost the same width…

> Also, thanks for the PR :)

Your welcome. :-)
Also, for a better localization, please take care of the number format by using Number.toLocaleString(_locale_) ([issue #360](https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/issues/360)). TIA.

franglais125:
@DevDef ok, agreed on all points! And yes, I do speak French, but use gnome in the English locale, hence why I don't notice these things you reported, merci encore.

As for the the `toLocaleString`... I was trying to prepare an overhaul of the locale work (checkout my branches). I'm a bit behind, and don't have that much time these days, but yes, it's something I want to fix eventually.

DevDef:
I’m glad to see that you are already taking care of the localized number format. :-)
Cheers,

Davy

Issue #359, opened 2017-04-16
GiB is not gettextized


DevDef:
Although MiB, KiB are gettextized and hence translated, the “GiB”, “kbps” and “'Mbps” strings remain in English (BTW, the [official abbreviations](https://en.wikipedia.org/wiki/Bit_rate) for bit rates are “kbit/s” and “'Mbit/s”, so they should be changed accordingly).

TIA,

Davy

franglais125:
Thanks for noticing this.

I added a fix in the localization PR I had already started ([e9ea188f85c6e48d26f66056bc0ee371b80b8858] in https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/pull/354)

DevDef:
Nice. Did you mean that gettextizing “GiB” string is just the first step?
I hope you’ll also replace “kbps” and “'Mbps” by “kbit/s” and “'Mbit/s”, or at least gettextized them.
The best, IMHO, would be to do both the replacement and the gettextization…

Issue #355, opened 2017-04-06
Unit texts are not aligned correctly after numerical values.


Ecron:
When clicking over the widget, the units shown after the numbers are not correctly aligned after the numerical values, as shown in the picture below (texts in Catalan):

![Screenshot](https://cloud.githubusercontent.com/assets/488803/24768266/46ef7940-1b02-11e7-9b8c-3bf835e250d0.png)

I guess the final result should be something like the following **mockup**:

![a](https://cloud.githubusercontent.com/assets/488803/24770863/b7d0d1a0-1b0b-11e7-8ffa-de2c96e324a8.png)

In this mockup, I also aligned the single-valued parameters to the first left column.

Is there any way to fix this layout?

franglais125:
Hey, thanks for opening this. So I fixed this in the link [1], but not exactly in the same way you are proposing.

Anyway, from your screenshot and mine (to be uploaded soon), I am not sure it looks much better, it looks a bit crowded. I won't open a PR for this. Instead I'll wait for more feedback. As I said, I'm not sure it looks better.

Thoughts anyone?

[1] https://github.com/franglais125/gnome-shell-system-monitor-applet/tree/align_units

Ecron:
In my opinion, that's a nice improvement, that's basically what I was asking for. I don't feel it more crowded, but more understandable, now every unit is next to its numerical value :+1:

In my mockup I aligned single-valued parameters to the first column, but that's just because I didn't understand why there's so much blank space between, for example, `CPU` and `7%`, or `Thermal` and `49ºC`.

DevDef:
Definitely better!

franglais125:
@Ecron do you mind posting a screenshot of the result with your patch?

Ecron:
Here it is:

![captura de pantalla de 2017-05-18 12-41-34](https://cloud.githubusercontent.com/assets/488803/26198722/6f4ed9a4-3bc7-11e7-9e75-4b3f5577501f.png)

franglais125:
Sure! It looks better in my opinion.

Moltes gracies!

franglais125:
@ecron (and everyone), what do you think o this mockup?

I added the percentage value for Memory and Swap, hence it looks more harmonised.

![screenshot from 2017-05-18 19-21-14](https://cloud.githubusercontent.com/assets/19195975/26227256/34692b9a-3bff-11e7-927c-acf118b5cdc9.png)

[...]

PR #369, opened 2017-04-30
Use kbit/s instead of kbps

franglais125:
Hi,

This PR is a proposal to change
- `kbps` to `kbit/s`
- `Mbps` to `Mbit/s`

as the units shown when using the option "Show network speed in bits". This was originally proposed by @DevDef in issue #359.

The second commit in this PR is simply updating the general file for translations, so that it's easier to track strings that need to be translated.

### DIFF
--- a/po/system-monitor.pot
+++ b/po/system-monitor.pot
@@ -253,6 +253,14 @@ msgstr ""
 msgid "MiB/s"
 msgstr ""
 
+#: extension.js:178 extension.js:1597 extension.js:1605
+msgid "kbit/s"
+msgstr ""
+
+#: extension.js:179 extension.js:1600 extension.js:1608
+msgid "Mbit/s"
+msgstr ""
+
 #: extension.js:1419 extension.js:1635
 msgid "MiB"
 msgstr ""
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -175,8 +175,8 @@ const smStyleManager = new Lang.Class({
     _diskunits: _('MiB/s'),
     _netunits_kbytes: _('KiB/s'),
     _netunits_mbytes: _('MiB/s'),
-    _netunits_kbits: 'kbps',
-    _netunits_mbits: 'Mbps',
+    _netunits_kbits: _('kbit/s'),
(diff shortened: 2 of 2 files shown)


franglais125:
I have added a fix for #360 . @DevDef Do you mind testing if this is ok, and if there are other numbers I might have missed?

DevDef:
There’s currently a bug in the GNOME 3.24 JavaScript back-end (mozjs38) when calling the Number.toLocaleString() method with no argument. What you did probably works on previous GNOME versions, but it doesn’t work for GNOME 3.24 (see [this discussion](https://github.com/jenslody/gnome-shell-extension-openweather/issues/184#issuecomment-296442405) and this [Mozilla bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=999003)).
The workaround is to get the current user’s locale and give it as an argument for the Number.toLocaleString() method (see [this patch](https://github.com/jenslody/gnome-shell-extension-openweather/issues/184#issuecomment-296442405) for gnome-shell-extension-openweather).

BTW, I wonder how we can translate the string “Display system informations in gnome shell status bar, such as memory usage, cpu usage, network rate…” (should rather be “Display system informations in **GNOME** shell status bar, such as memory usage, **CPU** usage, network rates…”, BTW) that is located in [metadata.json](https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/blob/master/system-monitor%40paradoxxx.zero.gmail.com/metadata.json) and that is displayed in the _Extensions_ tab of the GNOME Tweak tool.

Lastly, I completely agree that we need the fix you provided for the issue #355. ;-)

DevDef:
Please see my pull request #372 that fixes the issue with GNOME 3.24.

franglais125:
@DevDef Thanks for the checks! I cherry-picked and squashed your commits from the other PR in [bb4832f6921c5060e0118b18a3329591e681c4c1].

franglais125:
@chrisspen I removed the Locale fix from this PR, as it had conflicts with the other PR I opened.

I moved the fixes to https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/pull/375.

On this way, this PR does what the title says: it simply changes the units from kbps to kbit/s, etc.
[09bc07e4653cd4b4ef3a58950c9f0063fa0f388c] and [998f6f963514b8dd21e45a78cee2bde09bd37d9b]

PR #372, opened 2017-05-08
Use Number.toLocaleString() for better l10n

DevDef:
This patch increase the localization by displaying the numbers according to the user’s locale.
In order to be compatible with GNOME 3.24, the Number.toLocaleString() method has to be given the locale as an argument.


chrisspen:
Please merge recent commits to staging and resolve conflicts.

franglais125:
@chrisspen I think this PR was included in the big one (from me) you just merged.

@DevDef could you confirm? It's been a while and I don't remember 100%.

DevDef:
@franglais125 I think you’re right.

PR #386, opened 2017-08-16
Gpu load display

franglais125:
Adding support for Nvidia GPUs.

This uses the Nvidia `nvidia-smi` command, that is run async at the chosen interval.

1) The graph displays the load %, while the menu shows the load and the VRAM usage.
2) The menu display will look better once the "style" PR is merged as well.

Big thanks to @DavidLKing for the support and testing.


franglais125:
I've been researching on how to do this for AMD and Intel cards. Unfortunately, in both cases `sudo` privileges are required to run the respective commands. A way around this is to explain in the wiki how to add a policykit rule to allow for this without requiring the users' intervention.

In any case, I won't add that for now, but given enough interest we can perhaps take another look.

franglais125:
@chrisspen And I fixed this one too.

Snuggle:
Is there any chance this could be added to `gnome-system-monitor`?

Window's task manager is capable of showing GPU usage, but there's no alternative for Linux.

DavidLKing:
@Snuggle, you may want to consider notifying them: https://github.com/GNOME/gnome-system-monitor

This extension is independent of the gnome-system-monitor dev workflow.

Issue #400, opened 2017-10-01
Error and crash from battery functionality


NVieville:
Hello,

On Fedora 26, new applet launches well, but when display hits stand-by, re-activating the session doesn't show the applet.
This patch seems to fix this issue:

```
Date: Sun, 1 Oct 2017 16:53:59 +0200
Subject: [PATCH] Fix battery error and crash

---
 system-monitor@paradoxxx.zero.gmail.com/extension.js | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/system-monitor@paradoxxx.zero.gmail.com/extension.js b/system-monitor@paradoxxx.zero.gmail.com/extension.js
index a19a1f1..f8fbc4f 100644
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -1102,7 +1102,7 @@ const Battery = new Lang.Class({
         } else {
             displayString = this.percentage.toString()
         }
-        this.text_items[1].text = this.menu_items[3].text = displayString;
+        this.text_items[1].text = this.menu_items[0].text = displayString;
         this.text_items[0].gicon = this.gicon;
         this.vals = [this.percentage];
         this.tip_vals[0] = Math.round(this.percentage);
-- 
2.13.6

```

Cordially,

-- 
NVieville

NVieville:
This issue have been fixed by #403, then closing.

PR #403, opened 2017-10-04
Fix battery error and crash

NVieville:
Hello,

This pull request do exactly what was provided in the patch in issue #400.

Cordially,

-- 
NVieville

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -1102,7 +1102,7 @@ const Battery = new Lang.Class({
         } else {
             displayString = this.percentage.toString()
         }
-        this.text_items[1].text = this.menu_items[3].text = displayString;
+        this.text_items[1].text = this.menu_items[0].text = displayString;
         this.text_items[0].gicon = this.gicon;
         this.vals = [this.percentage];
         this.tip_vals[0] = Math.round(this.percentage);


chrisspen:
@franglais125 Can you confirm this? Do you know why the current index used in the code isn't causing more problems?

franglais125:
It's weird that we have this issue, and nobody complained earlier. I used the built-in battery indicator anyway, so perhaps I just didn't hit this problem.

We don't even seem to have that many indices for `this.menu_items[]`, just `0` and `1`. So I really don't know where `3` comes from.

Running `git blame` for the line we are changing takes me back to a commit from 2012, so many things might have changed since.

This being said, I'll check more carefully later today/tomorrow.

Thanks @NVieville for spotting this and for the PR!

NVieville:
@franglais125 You're welcome.
The difference was made by commit 8a9419440693b0eea9c2bde4b3bca2176609460e on 10 April 2017.
This one removed some "new St.Label()", in the create_menu_items function below line 1116, that led to have less items in "this.menu_items[]". Index 3 at line 1105 was here earlier, but from this commit it didn't point to an element of the "this.menu_items[]" array.
Glad to help.
Cordially,


-- 
NVieville

franglais125:
I can now confirm the issue and the fix.

I wasn't seeing this as I don't use the battery module. Well done @NVieville !

Here is the JS error from the log:
```
JS ERROR: TypeError: this.menu_items[3] is undefined
                                        Battery<._apply@/home/franglais125/.local/share/gnome-shell/extensions/system-monitor@paradoxxx.zero.gmail.com/extension.js:1105
```

chrisspen:
Thanks @franglais125

Issue #405, opened 2017-10-17
Menu text all jumbled when text scaling factor not set to 1


nturner:
I have (via dconf-editor) set org.gnome.desktop.interface.text-scaling-factor to 0.70 on my system. In the attached screenshot, the "System Monitor..." and "Preferences..." menu entries are the correct size, as are the numeric monitor values above. The label text (e.g. "Cpu") and unit text (e.g. "KiB/s Down") are, however, all different sizes, and don't line up with the numeric values.

![screenshot from 2017-10-16 20-09-40](https://user-images.githubusercontent.com/209285/31640866-e8d2fc66-b2ae-11e7-8949-23cb04767d72.png)

franglais125:
Thanks a lot for reporting this! I have a fix in a branch that I can't publish at the moment due to some permission issues with github. I'll repost soon!

franglais125:
Here we go, would you mind testing this branch https://github.com/franglais125/gnome-shell-system-monitor-applet/tree/font_sizes ?

I changed all styles from `px` to `em` so that sizes are relative and not absolute.

franglais125:
Just a note: this doesn't take care of the pie/bar charts for filesystems (that include labels such as "/home".

I'll look into it when I have some time, but it's not as easy to fix.

franglais125:
I added a new commit that fixes the graphs (both pie and bar styles).

franglais125:
This can be closed now, it's fixed in master.

PR #407, opened 2017-10-19
Fix font scaling

franglais125:
Hello all,

Here are two commits to fix the font scaling of the menu. Testing would be appreciated, in case I missed anything. But it's already much better, as we at least respond to scaling changes (requires session restart if you are testing).

Cheers


franglais125:
Addresses https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/issues/405.

PR #410, opened 2017-11-06
extension: make compatible with libnm

lkundrak:
libnm-glib has been deprecated in favor of libnm for ages. GNOME Shell
seems to be among the last bits to be ported. This is now being done [1]:

[1] https://bugzilla.gnome.org/show_bug.cgi?id=789811

Unfortunately, due to GLib limitations, libnm-glib.so and libnm.so can't
be loaded at the same time and the attempt to do so results in a nasty
crash. The extensions thus needs to be updated to libnm.

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -18,6 +18,10 @@
 
 // Author: Florian Mounier aka paradoxxxzero
 
+/* Ugly. This is here so that we don't crash old libnm-glib based shells unnecessarily
+ * by loading the new libnm.so. Should go away eventually */
+const libnm_glib = imports.gi.GIRepository.Repository.get_default().is_registered("NMClient", "1.0");
+
 let smDepsGtop = true;
 let smDepsNM = true;
 
@@ -40,7 +44,7 @@ const Me = ExtensionUtils.getCurrentExtension();
 const Convenience = Me.imports.convenience;
 const Compat = Me.imports.compat;
 
-let Background, GTop, IconSize, Locale, MountsMonitor, NMClient, NetworkManager, Schema, StatusArea, Style, gc_timeout, menu_timeout;
+let Background, GTop, IconSize, Locale, MountsMonitor, NM, NetworkManager, Schema, StatusArea, Style, gc_timeout, menu_timeout;
 
 try {
     GTop = imports.gi.GTop;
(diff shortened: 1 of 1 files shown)


franglais125:
Thanks a lot for this! I have yet to try it, my main concern is wrt backwards compatibility. Since I see you worked on this upstream: should we expect the extension to stop working on older setups?

I haven't worked on this extension since the beginning, so I can't speak for the whole span of it, but looking at the code at metadata.json, there is some effort to support *very* old Gnome-Shell versions.

If a new release includes this new codepath (`NM` instead of `NMClient`), do we need to drop older Gnome versions? And if so, how old? Or is this to be expected for the 3.28 release? I'm not exactly sure on how this is managed upstream, as I see the NetworkManager component is in the 1.11dev state (looking at https://cgit.freedesktop.org/NetworkManager/NetworkManager/log/).

Cheers, Fran

lkundrak:
Hello,

> Thanks a lot for this! I have yet to try it, my main concern is wrt backwards compatibility. Since I see you worked on this upstream: should we expect the extension to stop working on older setups?
>
> I haven't worked on this extension since the beginning, so I can't speak for the whole span of it, but looking at the code at metadata.json, there is some effort to support very old Gnome-Shell versions.

Well, the patch here keeps compatibility with both libnm and libnm-glib. It works with current as well as future versions of the shell.

> If a new release includes this new codepath (NM instead of NMClient), do we need to drop older Gnome versions? And if so, how old? Or is this to be expected for the 3.28 release?

Not necessarily. The only reason to drop compatibility with the old version is supporting both becomes a maintenance burden.

In the even that happens, there's no real good way to do that at the moment, since GNOME Shell 3.22.0 defaults to disabling the version checks altogether [1].

[1] https://git.gnome.org/browse/gnome-shell/commit/?id=5e0e3edc7

We could in future change the Shell to only allow extensions whose oldest "shell-version" is not newer than the current version but that would break shells 3.22-3.26, so I'm not sure it's an option.

> I'm not exactly sure on how this is managed upstream, as I see the NetworkManager component is in the 1.11dev state (looking at https://cgit.freedesktop.org/NetworkManager/NetworkManager/log/).

The upstream NetworkManager development has very little to do with this. The "new" API is present many long years (since NetworkManager 1.0) -- the only somewhat relevant question is when GNOME Shell switches. I hope it's going to be 3.28. The patched version of the extension should ideally happen before that.

franglais125:
I have now tested this, working fine over here.

@chrisspen do you mind pulling this?

PR #415, opened 2017-11-22
Fix empty St.Label() calls

NVieville:
Hello,

There seems that empty labels crash gnome-shell 3.26.x, and as I hit this issue on Fedora 27, and I don't know yet what the sources of this crashes are, I patched extension.js in order to fix it. Unfortunately this correction didn't fix gnome-shell crashes. Maybe the issue is elsewhere.

See https://bugzilla.redhat.com/show_bug.cgi?id=1469129 for more about the crashes.

Cordially,

-- 
NVieville

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -907,7 +907,7 @@ const ElementBase = new Lang.Class({
             let tipline = new TipItem();
             this.tipmenu.addMenuItem(tipline);
             tipline.actor.add(new St.Label({text: _(this.color_name[i])}));
-            this.tip_labels[i] = new St.Label();
+            this.tip_labels[i] = new St.Label({text: ''});
             tipline.actor.add(this.tip_labels[i]);
 
             this.tip_unit_labels[i] = new St.Label({text: unit[i]});
@@ -1132,6 +1132,7 @@ const Battery = new Lang.Class({
                 gicon: Gio.icon_new_for_string(this.icon),
                 style_class: Style.get('sm-status-icon')}),
             new St.Label({
+                text: '',
                 style_class: Style.get('sm-status-value'),
                 y_align: Clutter.ActorAlign.CENTER}),
             new St.Label({
@@ -1143,6 +1144,7 @@ const Battery = new Lang.Class({
     create_menu_items: function () {
         return [
             new St.Label({
+                text: '',
                 style_class: Style.get('sm-value')}),
             new St.Label({
                 text: '%',
@@ -1310,6 +1312,7 @@ const Cpu = new Lang.Class({
     create_text_items: function () {
         return [
             new St.Label({
+                text: '',
                 style_class: Style.get('sm-status-value'),
                 y_align: Clutter.ActorAlign.CENTER}),
             new St.Label({
@@ -1320,6 +1323,7 @@ const Cpu = new Lang.Class({
     create_menu_items: function () {
         return [
             new St.Label({
+                text: '',
                 style_class: Style.get('sm-value')}),
             new St.Label({
                 text: '%',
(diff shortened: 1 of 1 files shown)


franglais125:
Thanks for this patch, but I have a couple of questions:

1. It would perhaps be better to open an issue regarding what this patch fixes, so that there is more information about it. For instance, providing an error log, etc.
2. What version of the extension are you using when you have the crashes? A new version was uploaded only a few days ago, but it's still unapproved. Can you confirm that you still have this problem when running the latest version from `git` (if you aren't already)?
3. In the bug report you linked, I've only seen one user listing this extension. Are you sure this extension is to fault?

I'm not saying there is no issue here, I'm just trying to better document what it is, and how we react to it. I, for one, have been using 3.26 on Debian for quite some time and haven't seen this sort of crashes.

Cheers,
Fran

NVieville:
Hello,

Sorry for not being accurate in my previous message.

1. I didn't open an issue as I'm not certain this extension is the cause of the gnome-shell crash. The error log I was trying to fix was plenty of these messages in /var/log/messages (or with "journalctl /usr/bin/gnome-shell" command):

gnome-shell[1803]: invalid uninstantiatable type '(null)' in cast to 'ClutterText'
gnome-shell[1803]: clutter_text_get_editable: assertion 'CLUTTER_IS_TEXT (self)' failed
gnome-shell[1803]: clutter_text_get_text: assertion 'CLUTTER_IS_TEXT (self)' failed
gnome-shell[1803]: clutter_text_set_text: assertion 'CLUTTER_IS_TEXT (self)' failed

2. The version of the extension I'm running correspond to 61b0a60 commit from this repository.

3. As I said in first point, I'm not confident in saying this extension is the cause of the gnome-shell crashes, but as one of the cause seems to be empty string labels, and as this extension is one of the two I uses (the other is "removable drive menu"), I thought it wouldn't hurt to fix empty labels.

I'm afraid the bug report mentioned above has multiple causes, and probably this extension is not one of them.

Thanks for your comments.

Cordially,


-- 
NVieville

franglais125:
Thanks a lot for the (3) clarifications!

>    gnome-shell[1803]: invalid uninstantiatable type '(null)' in cast to 'ClutterText'
>    gnome-shell[1803]: clutter_text_get_editable: assertion 'CLUTTER_IS_TEXT (self)' failed
>    gnome-shell[1803]: clutter_text_get_text: assertion 'CLUTTER_IS_TEXT (self)' failed
>    gnome-shell[1803]: clutter_text_set_text: assertion 'CLUTTER_IS_TEXT (self)' failed

I've certainly seen these error in my logs every once in a while. Any idea on how to quickly trigger this, just to confirm?

I'll give your fix a try and check for this problem, again, a bit subjectively as I'm not sure when this appears.

Thanks for packaging this :)

NVieville:
> I've certainly seen these error in my logs every once in a while. 
> Any idea on how to quickly trigger this, just to confirm?

For the moment, I can't say. Still investigating it.

> I'll give your fix a try and check for this problem, again, a bit subjectively as 
> I'm not sure when this appears.

Sorry but I'm in the same situation, trying to catch it, so same point of view.

> Thanks for packaging this :)

You're welcome.

NVieville:
Hello,

After few hours using gnome-shell and this extension with the proposed fix, I can confirm that clutter messages in gnome-shell logs are gone, and that gnome session stopped to crash.
One must note too that to get the benefit of this fix one have to close session, restart gdm and open new session (or reboot). Alt-f2 + r doesn't stop clutter messages in logs and gnome session crash.

Cordially,


-- 
NVieville

NVieville:
Hello,

I've made some modifications on the initial pull request. I confirm that adding systematically one "text" field (even empty) for each call to "St.Label()" remove all the:

gnome-shell[6584]: invalid unclassed pointer in cast to 'ClutterText'
gnome-shell[6584]: clutter_text_get_editable: assertion 'CLUTTER_IS_TEXT (self)' failed
gnome-shell[6584]: clutter_text_get_text: assertion 'CLUTTER_IS_TEXT (self)' failed

entries seen with "journalctl /usr/bin/gnome-shell" command.

Cordially,


-- 
NVieville

franglais125:
Fwiw, I've been using this patch locally ever since, and all warnings have disappeared.

franglais125:
Merged [447d2fb81364144f4ddabe7f51ac3e3be26188b9].

Thanks a lot for the fix provided!

Issue #419, opened 2017-12-09
Arch Linux: Extension crashes after lock screen


dkumor:
The extension loads fine after login, but after leaving computer idle until it goes to lockscreen, and unlocking, the extension is gone.

The following show up in journalctl log for gnome-shell
```
System monitor applet disable
System monitor applet enabling
Extension "system-monitor@paradoxxx.zero.gmail.com" had error: Error: Object Gio.Settings (0x564de4bfe520), has been already deallocated - impossible to access to it. This might be caused by the fact that the object has been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs
```

I am using the AUR version of the extension: https://aur.archlinux.org/packages/gnome-shell-extension-system-monitor-git/
This issue showed up recently, and upgrading to most recent AUR version did not help. 

I am using a fully updated system (Gnome 3.26.2), the issue has persisted through multiple reboots.

Any suggestions?

FSMaxB:
@dkumor: Try downgrading gjs to 1.50.2, this might be related to https://gitlab.gnome.org/GNOME/gjs/issues/21

Essentially the GNOME folks changed gjs in such a way, that it crashes when destroyed objects are used from JavaScript from my understanding.

franglais125:
@FSMaxB Thanks a lot for the tip. I've seen very similar errors suddenly being reported on various extensions. It seems that a few fixes will need to be applied here and there,

dkumor:
@FSMaxB I can confirm, downgrading gjs to `gjs-1.50.2-1` seems to have fixed the problem based on an initial test.

For clarity, this was the command used:
```
sudo pacman -U /var/cache/pacman/pkg/gjs-1.50.2-1-x86_64.pkg.tar.xz
```

Since this is a workaround, should I leave this issue open?

franglais125:
@dkumor I'd leave it open, as eventually this will affect more users. I'll try to tackle this once the newer package is available on Debian. If someone else can debug this, then we can perhaps find the source of the problem sooner.

Following the link to the gitlab issue, it seems that the crash will be "downgraded" to a warning relatively soon, so perhaps when that happens it will be easier to spot the error (and nicer for users anyway!).

auipga:
the workaround by @dkumor doesn't work for me. I tried `gjs-1.50.2-1` and `gjs-1.50.1-1`

noreun:
[...]

PR #426, opened 2018-01-31
Fix missing Schema object when unlocking the screen

franglais125:
Hopefully addresses #419

The problem was that the Schema was created in `init()`, and then destroyed in `disable()`.

I moved the creation to `enable()` so that it gets created properly when needed, i.e. with each new session.


franglais125:
Merged [91972278940b157bb6554429c5b61ba5cf333e3e].

PR #431, opened 2018-02-14
Cpu frequency enhancements

roadrunner2:
This contains a few fixes and enhancements to the cpu-frequency graph display, addressing two existing tickets in the process. See the individual commits for details.

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -1515,15 +1515,17 @@ const Freq = new Lang.Class({
     elt: 'freq',
     item_name: _('Freq'),
     color_name: ['freq'],
-    _init: function () {
-        this.item_name = _('Freq');
+    cpuid: 0,
+    _init: function (cpuid) {
+        this.item_name = _('Freq') + (cpuid < 0 ? '' : ' ' + (cpuid + 1));
+        this.cpuid = cpuid;
         this.freq = 0;
         if (GLib.file_test('/sys/devices/system/cpu/cpu0/cpufreq/scaling_cur_freq', 1 << 4)) {
             this.fixed_freq = -1;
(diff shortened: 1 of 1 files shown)


roadrunner2:
Apologies, the first version of the commits contained some lint errors - all fixed now.

franglais125:
I'll test this when I get some time, at least to provide some feedback.

Quick comment: for someone who has more than one core (which I assume is the vast majority of users), there will be now one graph *per* core? It'd perhaps be better if this could be optional (like in the Cpu element). For someone having a quad-core or more it could be too much.

Which bring me to the 2d question: what were we doing so far? The frequency displayed was the average over all the cores?

Feel free to educate me if my assumptions/understanding above are wrong. And thanks for your work on this :)

roadrunner2:
> Quick comment: for someone who has more than one core (which I assume is the vast majority of users), there will be now one graph per core? It'd perhaps be better if this could be optional (like in the Cpu element). For someone having a quad-core or more it could be too much.

Apologies for not making this clear: there is an existing preference, under the cpu tab, labelled "Display Individual Cores": currently this only affects the cpu usage graph, but with the last commit above it also affects the cpu frequency graph. I.e. If this preference is off, then a single frequency graph is displayed like now, and if the preference is on then one frequency graph per cpu (like the cpu usage graphs) is displayed. I figured it made more sense to re-use this preference, but am open to creating a separate one just for the frequency graph(s).

> Which bring me to the 2d question: what were we doing so far? The frequency displayed was the average over all the cores?

Correct. Note that there is also pull request #352 which attempts to improve this single graph.

return42:
@roadrunner2  can you merge this PR with current master / Thanks!

roadrunner2:
> @roadrunner2 can you merge this PR with current master / Thanks!

Done.

return42:
See https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/issues/496#issuecomment-466770161

chrisspen:
Sorry to keep asking you to re-merge master, but could you resolve the current conflicts?

Issue #202, opened 2013-02-18
Regular "micro-freezes" when extension is enabled in gnome-shell 3.6.3


adlorenz:
Just upgraded Fedora to release 18 which comes with gnome-shell 3.6.3 and noticed annoying issue when system-monitor extension is enabled. The desktop suffers from "micro-freezes" which take just split-second yet are utterly annoying (ie. mouse or typing cursor chokes, often leaving unwanted effect). 

I've managed to nail the problem down to system-monitor, as these freezes stop happening when extension is disabled. Also, the freezes seem to occur in regular intervals, approx. every 4-5 seconds. I've tried to adjust extension settings and remove lesser used sensors, leaving just CPU and MEM sensors active, yet the problem didn't stop occurring.

rzippert:
I am experiencing this screen freeze issue but it's not related to any network or disk mounts. In my case it's related to reading the fan speeds with i8k and its been happening since I was using Fedora 19 (I'm now on 21). The freezes occur exactly in the time I set the fan readings to update. I've already removed i8k, however its option doesn't disappear from the fan drop down menu and, also, didn't stop the freezes. I'm using a Dell Vostro 2450 laptop with Fedora 21, so I'm not sure if i8k was supposed to work with it, but still, this is what causes the problem here. I'm now looking for some way to disable these fan readings definitely...

aspexi:
Hi guys

I had this issue for a long time and I couldn't figure out what causes this.

Happily I have found this topic. 

Currently I have:

Ubuntu 14.10
Gnome Shell 3.14.3
the latest gnome-shell-system-monitor-applet according to extensions.gnome.org

The issue occures.

Anyway, I've just downloaded the latest master codebase and compared it to my latest codebase. These files are different! (I have not modified them)

I've replaced all files with master and now the issue is gone.

The question is why extensions.gnome.org shows that I have the latest version of the plugin when I have not!?

rzippert:
As an update from the last message (1 january), tested again and no more freezes. No idea what has changed...

laurentbriet:
I'm experiencing micro freeze on a frequency around a second or two after updating to ubuntu 17.10 from 17.04. Disabling the extension and the freezes are gone.

xiaosong0911:
Micro freezes happen in every several minutes, with archlinux and gnome 3.26.2 in my laptop. Disabling this extension there is no problem.

EvgeniiFrolov:
[...]

PR #443, opened 2018-03-21
Support GNOME 3.28.

marxin:
I've just tested the version and it works fine.

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/metadata.json
+++ b/system-monitor@paradoxxx.zero.gmail.com/metadata.json
@@ -1,5 +1,5 @@
 {
-    "shell-version": ["3.4","3.6","3.8","3.10", "3.12", "3.14", "3.16", "3.18", "3.20", "3.22", "3.24", "3.26"],
+    "shell-version": ["3.4","3.6","3.8","3.10", "3.12", "3.14", "3.16", "3.18", "3.20", "3.22", "3.24", "3.26", "3.28"],
     "uuid": "system-monitor@paradoxxx.zero.gmail.com",
     "name": "system-monitor",
     "url": "https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet",


franglais125:
Thanks!

gugaiz:
Are you going to merge this? Ubuntu 18.04 is about to see the light and this extension is not listed as supported on 3.28.

franglais125:
It was already merged.

We need to upload a new version to extensions.gnome.org. I'll probably need to ping @chrisspen or a reviewer.

chrisspen:
@franglais125 Is it ready for a release? I try not to upload too often, because it seems like a lot of bugs slip through.

franglais125:
The new "release" is not really that big. We only had 9 commits since v33...

That being said, the quickest approach to support Gnome 3.28 (not just Ubuntu) is to upload a new version with the metadata modification **only**. For instance, I did just that for another extension I maintain. By doing so ,we avoid getting stuck on the review list, which can take a long time these days.

Modifications to metadata and translation files get auto-approved.

See my example from today: https://extensions.gnome.org/review/8040

piegamesde:
Only updating the metadata is not enough IMO. There were a few bugs fixed since then (the tooltips, the log warnings, the frequency reading, the spacing, ...) that shouldn't be ignored.

chrisspen:
@franglais125 [This user](https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/issues/202#issuecomment-378898549) reported that the current git master is unusable on his platform.

franglais125:
@piegamesde : Ideally we wojld of course include all the fixes, I'm not against that. What I'm proposing *for now* is to make a release that merely advertises 3.28 support. Such releases get automatically approved. If we include any other fix, we will get stuck in the review queue for days/weeks/months (take a look here extensions.gnome.org/review). The goal here, frlm my perspective, is to give access to this extension to people running 3.28 as soon as possible.

Once this is done, we can *then* look at making a new release with actual code modifications: features, fixes, changes,etc. Now this next version *will* indeed go through the review queue, as usual.

----------------------
@chrisspen : I hope the above ecplanation addresses your concern as well. The new release I propose (until we are ready for an "actual" release) won't change anything for that user, nor for anyone else, as only the metadata will have changed (adding 3.28 support). I'll comment on that separate issue soon, not to conflate everything here.

chrisspen:
@franglais125 I take your point. I agree, it's a good idea to make a release just to change some metadata that will update gnome support. However, that would require pushing up master, right? Have the commits made to master since the last release caused the problems in issue 202? If they have, even a release to change metadata would introduce other changes. The solution to this is to adopt a new branching strategy that would keep all merged fixes in a dedicated "develop" branch, so we can push up minor fixes to master without deploying other untested changes.

p91paul:
@chrisspen if there isn't a tag of the last released version on e.g.o. you can just download the zip file from e.g.o., edit metadata.json and reupload it.

piegamesde:
As far as I know, #202 does affect almost all users, not only one person. The fix provided by @franglais125 is still not in the master. Also, the fix has a bug that consumes file pointers until Gnome crashes and restarts, which is way worse than the original bug.

franglais125:
I tagged v35: https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/releases/tag/v35

With respect to the latest upload, this only contains a modification to the metadata file.

chrisspen:
@franglais125 I've uploaded your tag for review.

Issue #440, opened 2018-03-19
gpu utilization graph is empty


mahmoodn:
I have installed gnome-shell-system-monitor-applet on ubuntu 17.10. The nvidia driver has been installed and I can see the gpu utilization from the nvidia-smi. However, the applet shows nothing.
How can I debug to see what is the problem?

piegamesde:
I just tried it out and it seems to work fine. What hardware are you using? Can you try out the script (`gpu_usage.sh`) in the installation directory and tell its output?

franglais125:
You might also need to restart the session to make sure it's working.

Otherwise, you can try what @piegamesde suggested, and look for errors on `sudo journalctl /usr/bin/gnome-shell` that could be relevant.

mahmoodn:
I have noticed that when I restart the machine, the gpu graph is fine. However, after some minutes the chart becomes empty. Previously I thought that the problem is related to screen lock (lock after some minutes). But, now I see it fails even without screen lock.

Please see the picture as below. As you can see the gpu_usage.sh is fine!
https://pasteboard.co/HcEHJDm.png

mahmoodn:
I have noticed that when I restart the machine, the gpu graph is fine. However, after some minutes the chart becomes empty. Previously I thought that the problem is related to screen lock (lock after some minutes). But, now I see it fails even without screen lock.

Please see the picture as below. As you can see the gpu_usage.sh is fine!
https://pasteboard.co/HcEHJDm.png

piegamesde:
What if you enable the digit display? What does it show?

mahmoodn:
Yes it also freezes at the last value.

cbcoutinho:
I'm having this same issue, at startup the GPU meter seems to follow the output of `nvidia-smi` but pretty quickly it either 

a) goes to 0-1% even when GPU usage is higher
b) gets fixed at some value it was at (e.g. 30% and doesn't budge 
c) goes to NaN

Restarting the window manager helps for a few minutes but doesn't solve the problem

franglais125:
Thanks for reporting. Unfortunately I don't own an Nvidia GPU, so it's really hard to debug. One thing I can do is provide a debug branch, and then ask for logs. Give me some time for this. Anyone else pitching in would be appreciated obviously!

cbcoutinho:
[...]

PR #451, opened 2018-04-23
Update gpu_usage.sh

Seranth:
nvidia-smi doesn't seem to work very well for people, nvidia-settings seems more reliable. For me the following change grabbed the gpu utilization from nvidia-settings instead of from nvidia-smi and it works great.

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/gpu_usage.sh
+++ b/system-monitor@paradoxxx.zero.gmail.com/gpu_usage.sh
@@ -25,4 +25,4 @@
 nvidia-smi -i 0 -q -d MEMORY | grep -A4 -i gpu | egrep -i "used|total" | awk '{print $3}'
 
 # This line will print the GPU usage in %.
-nvidia-smi -i 0 -q -d UTILIZATION | grep Gpu | awk '{print $3}'
+sed -e 's#.*=\(\)#\1#;s'/.$//'' <<< $(nvidia-settings -q all | grep -i GPUUtilization | grep graphics | awk '{print $4}')


kgreenek:
I verified that this works on Ubuntu 18.04 with an nvidia gtx 780

kriegalex:
For me the gpu_usage.sh works fine with Ubuntu 18.04, but it is somewhere inbetween that the info is lost to display on the GUI

franglais125:
@DavidLKing would you mind trying this patch? I remember you helped me set up the feature. I don't personally own a NVIDIA gpu to test it.

@kriegalex you have this problem with or without this patch? Or in all cases?

kriegalex:
> you have this problem with or without this patch? Or in all cases?
I use the default system-monitor, so the non patched gpu_usage.sh. When I manually execute it, it works fine, meaning *nvidia-smi* works well on my laptop. But then on the GUI the informations don't display correctly (always 0% usage and 0 memory used/total).

Maybe I should open a separate issue.

DavidLKing:
@franglais125 alas I'm in Europe at the moment and my usual GPU rig is running in server mode in the States! All the other rigs I use with GPUs are in clusters or the cloud and I won't be back until August :(

kgreenek:
It worked at first, but now I'm seeing the GUI not be udpated. This patch fixes the gpu_usage.sh script. Running it prints the correct gpu usage. But the GUI is showing 0% now. So perhaps that is a separate issue.

franglais125:
@kgreenek There is already an open issue about it: #440 

@DavidLKing No worries! Enjoy your trip.

PR #453, opened 2018-04-26
Update extension.js

Seranth:
This change allows you to open the system monitor in Ubuntu 18.04.


franglais125:
While this solves the problem for Ubuntu, this patch breaks the lookup for everyone else. Could you rework it so that it works for everyone?

return42:
I'am using Ubuntu 18.04.2 LTS and cant see any problems / can we reject this PR?

chrisspen:
I agree. I'm not sure what problem this fixes. I'm running Ubuntu 18 and I can open system monitor as is.

PR #476, opened 2018-10-18
scale width of elements if compact display is on

tonylambiris:
I made this from my initial confusion as to what `Compact Display` did in the preferences.

For years now I was always under the impression that this option would sort of "shrink" elements down -- and it did to a degree just not one that I was able to notice but of course became more clear while I was digging into the code.

Anyway, I hope you find this useful/handy. It's especially useful for me since I have a work laptop that does has a UHD display, a desktop setup with two 4k monitors, an older Acer laptop that has a max resolution of 1920x1080, etc.

Regardless of the resolution I find that I my menu-bar real estate runs out fairly quickly due to the all the extensions I have running and it's a much easier/faster solution than manually updating all the widths of all the elements.

Thanks!


chrisspen:
Could you please resolve conflicts?

tonylambiris:
@chrisspen sure sorry just seeing this now.

tonylambiris:
@chrisspen should be all set now, thanks!

chrisspen:
Thanks.

PR #486, opened 2018-11-16
Update extension.js

wolfgangwuerfl:
Fix deprecated GenericContainer, use ClutterActor

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -656,7 +656,8 @@ const TipMenu = new Lang.Class({
     _init: function (sourceActor) {
         // PopupMenu.PopupMenuBase.prototype._init.call(this, sourceActor, 'sm-tooltip-box');
         this.parent(sourceActor, 'sm-tooltip-box');
-        this.actor = new Shell.GenericContainer();
+        //this.actor = new Shell.GenericContainer();
+        this.actor = new Clutter.Actor(); // <<< Fix deprecated Shell.GenericContainer 
         this.actor.connect('get-preferred-width',
             Lang.bind(this, this._boxGetPreferredWidth));
         this.actor.connect('get-preferred-height',


wolfgangwuerfl:
With Gnome Shell 3.31, GenericContainer has been removed

jrevillard:
I'm facing the same issue, could this be merged ?

jrevillard:
I have updated accordingly but this seems not enough to make it working with gnome-3.31.. Anything else to update ?

jrevillard:
```
nov. 20 13:29:35 ordiwork gnome-shell[11906]: System monitor applet enabling
nov. 20 13:29:35 ordiwork gnome-shell[11906]: Extension "system-monitor@paradoxxx.zero.gmail.com" had error: Error: No signal 'get-preferred-width' on object 'ClutterActor'
```

wolfgangwuerfl:
Maybe i missed something, comment out for testing the lines below with:
get-preferred-width
and 
get-preferred-height 

and if it complains, the line with 

allocate

jrevillard:
Exactly, I need to comment the 3 properties and it works, thxs!

chrisspen:
Why is the old code left in but commented out? If it's still necessary for older Gnome versions, please update the code to work with them. We don't want the extension to break for legacy systems.

return42:
Tested backward compatibility of `Clutter.Actor` from this PR with GNOME Shell 3.28.3 and got `Error: No signal 'get-preferred-width' on object 'ClutterActor'`

return42:
looked at [Deprecate Shell.GenericContainer](https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/153): 

> Shell.GenericContainer exposes the size negotiation machinery through the use of signals.

In `Clutter.Actor` these signals do not exists and we have to use the counterpart function e.g. [allocate](https://developer.gnome.org/clutter/stable/ClutterActor.html#clutter-actor-allocate).  What I mean: this PR is broken and has never been tested with GNOME-Shell `V>3.30`.

Can we reject this PR?

Anyway a solution for [the depreciated Shell.GenericContainer](https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/153) is needed.

chrisspen:
Merging #510 may have fixed some of these issues, but it's also created conflicts with this PR. Could you please resolve these so we can re-test your changes?

chrisspen:
Can you please resolve conflicts?

PR #490, opened 2018-12-04
Bugfix/fix net mount related login hangs after susped

### LINKED ISSUE #350: Problems with relogin after sleep/screensaver in Gnome when extension is enabled
When I used this app, I sometimes had to wait 10-15 seconds after I put in my password after sleep/screensaver. Sometimes, even the password screen wouldn't vanish at all and I had to restart X.  If the login was successful (after the ~10sec wait),  the bar charts were empty / just starting to get new values (not sure if this is by design: https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/issues/329). This happened on two Laptops I set up (ubuntu 16.10 + debian stretch).

I tried to debug this but couldn't find anything suspicious in the journal. The problem went away after I disabled the system monitor extensions.

Do you have any way how I can get more information out of the extension?

koenlek:
Fixes #350 

See `FIXME` note in function `is_ro_mount` about the actual issue, which is now circumvented with `ENABLE_NETWORK_DISK_USAGE` set to `false`.

To reproduce the issue on original extension version:
1. Setup a VPN connection (I used OpenVPN)
2. Make a network mount over VPN (I used `sftp://username@server/home/username`)
3. Suspend system, then wake up system, this breaks the VPN connection.
4. Try to login, it hangs, because `is_ro_mount` hangs on the broken network mount. I verified this by printing log statements at start and end of that function.

I fixed this by swapping the order of an "if" check: first check if its a network mount and if so, skip it, if its no network mount, then check if its a RO mount (is_ro_mount).

This might also reproduce the issue:
1. Mount a network mount. 
2. Disconnect network (e.g. unplug or disable wifi). Now the mount is 'broken'.
3. Suspend system, then wake up system. OR: just lock the screen (win+L) and login again.
4. Try to login, it probably hangs too.

I also made the logging statements more clear and consistent, to make it easier to grep them out of the journalctl output.

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -424,8 +424,8 @@ const smMountsMonitor = new Lang.Class({
         }
         let mount_lines = this._volumeMonitor.get_mounts();
         mount_lines.forEach(Lang.bind(this, function (mount) {
-            if (!this.is_ro_mount(mount) &&
-                (!this.is_net_mount(mount) || ENABLE_NETWORK_DISK_USAGE)) {
+            if ((!this.is_net_mount(mount) || ENABLE_NETWORK_DISK_USAGE) &&
+                 !this.is_ro_mount(mount)) {
                 let mpath = mount.get_root().get_path() || mount.get_default_location().get_path();
                 if (mpath) {
                     this.mounts.push(mpath);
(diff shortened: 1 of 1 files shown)


koenlek:
@chrisspen : did you get any chance to look at this? It's a very simple fix, but I'm sure it will solve issues of many others with me that use network mounts over VPN.

koenlek:
@chrisspen or @franglais125, could you review this PR? It's small and very useful, if you ask me ;)

arigit:
+1 I have tested this for two weeks - the extension crashes are gone. Please merge :)

koenlek:
@chrisspen : thanks for merging. I see its not officially released yet though. When can we expect it to be released (such that I can recommend using to my colleagues :wink:)?

PR #500, opened 2019-02-17
Use Gio.Subprocess for cleaner child process handling

RenWal:
I'd like to propose this improvement on #461 to get rid of the GLib invocation in the GPU monitor.

While creating temporary streams to encapsulate and then close the file descriptors is certainly a way to do it, I think moving away from direct GLib calls is a better solution.

`Gio.Subprocess` allows for cleaner resource management as it handles all the process io file descriptors itself. Thus, the extension code only has to deal with the output data and never with the file descriptors, reducing the amount of code required.


RenWal:
This should close Launchpad bug #1776916 and GitHub issue #440

RenWal:
It seems to me that when `create_menu_items()` is called initially, this is before `refresh()` has necessarily completed its asynchronous call, meaning `useGiB` is un-initialized. Also, the current implementation doesn't update the menu items when the useGiB flag is updated. This should be fixed in this commit.
(Sorry for the force-push, I messed up with the linter)

chrisspen:
Could you please re-merge this with master?

RenWal:
Done

chrisspen:
Thanks.

chrisspen:
I'm running your branch locally. I'll test it for a week and I'm merge it if everything looks good.

PR #503, opened 2019-02-22
(v2) Makefile: complete rewrite with target 'help'

return42:
This is v2 with suggestions from @brianjmurrell (v1 see #499).

````
$ make help
Install or remove (and reload) of the extension, for the local user
or as admin for all users:

  make [install|remove]                        # for the local user
  make DESTDIR=/ [install|remove] clean        # as admin for all users

Use environment VERSION=n.m to set verison string in the metadata and in
the generated zip-file explicit.  If no VERSION is passed, the current
commit SHA1 is used as version number in the metadata while the generated
zip file has no string attached.

Other targets are:

  zip-file  - build and zip ./system-monitor@paradoxxx.zero.gmail.com.zip
  reload    - reload extension system-monitor@paradoxxx.zero.gmail.com
  clean     - remove most generated files
  extension - rebuild schemas/gschemas.compiled
  translate - generate translation from po/ files

control verbosity:

  make V=0 [targets] -> quiet build (default)
  make V=1 [targets] -> verbose build
````


chrisspen:
Thanks.

Issue #508, opened 2019-03-09
TypeError: this._construct is undefined


alephnull:
Causes the extension to not work. Seen on Ubuntu Disco Dingo. Doesn't make a difference if it is installed via the repo or the website.

In case it is relevant, the following was seen in journalctl just before the error.

Some code accessed the property 'ENABLED_EXTENSIONS_KEY' on the module 'extensionSystem'. That property was defined with 'let' or 'const' inside the module. This was previously supported, but is not correct according to the ES6 standard. Any symbols to be exported from a module must be defined with 'var'. The property access will work as previously for the time being, but please fix your code anyway.

This extension is pretty awesome. Longtime iStats user here.

shemgp:
This might be how it was fixed in dash to dock: https://github.com/micheleg/dash-to-dock/pull/881/commits/6382e38b195781f0bf89a8d9b0edc8d4fa387e45 so it may apply here also.

dino99:
Ubuntu 19.04 GS 3.32

 Extension "system-monitor@paradoxxx.zero.gmail.com" had error: TypeError: this._construct is undefined
                                                    0 check_sensors() ["/usr/share/gnome-shell/extensions/system-monitor@paradoxxx.zero.gmail.com/prefs.js":60]
                                                    1 add() ["/usr/share/gnome-shell/extensions/system-monitor@paradoxxx.zero.gmail.com/prefs.js":227]
                                                    3 _init/<() ["/usr/share/gnome-shell/extensions/system-monitor@paradoxxx.zero.gmail.com/prefs.js":362]
                                                    6 _init() ["/usr/share/gnome-shell/extensions/system-monitor@paradoxxx.zero.gmail.com/prefs.js":318]
                                                    10 buildPrefsWidget() ["/usr/share/gnome-shell/extensions/system-monitor@paradoxxx.zero.gmail.com/prefs.js":377]

sibradzic:
Ubuntu 19.04 here, same issue. This may be actually be gnome-shell issue or missing gir1.2 stuff?
```
 3月 11 18:48:12 A285 gnome-shell[2279]: JS WARNING: [resource:///org/gnome/gjs/modules/_legacy.js 67]: reference to undefined property "_construct"
 3月 11 18:48:12 A285 gnome-shell[2279]: Extension "system-monitor@paradoxxx.zero.gmail.com" had error: TypeError: this._construct is undefined
```
[...]

Issue #505, opened 2019-02-25
v36 does not support GS 3.31.90 (3.32)


dino99:

dino99:
ubuntu 19.04
Have upgraded GS from 3.30.2 to 3.31.90, and that breaks system-monitor: all icons gone.
So waiting an upgrade supporting the new coming Gnome-Shell 3.32

xinke2411:
I have same problem on ubuntu 19.04 with gnome-shell 3.31.90 .

Hentioe:
Gnome-Shell 3.32 has been released and is no longer compatible.

PR #510, opened 2019-03-14
Make it work with gnome-shell 3.32

shemgp:
Finally after two nights, it works with the new gnome-shell 3.32:
* using ES6 classes (fixes: #508)
* `Shell.GenericContainer` converted to `Clutter.Actor` (#486)
* probably not compatible with non-ES6 gnome-shell

Evidence:
Screenshot from 2019-03-15 00-07-54
Screenshot from 2019-03-15 00-08-06

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -20,31 +20,31 @@
 
 /* Ugly. This is here so that we don't crash old libnm-glib based shells unnecessarily
  * by loading the new libnm.so. Should go away eventually */
-const libnm_glib = imports.gi.GIRepository.Repository.get_default().is_registered('NMClient', '1.0');
 
-let smDepsGtop = true;
-let smDepsNM = true;
+var libnm_glib = imports.gi.GIRepository.Repository.get_default().is_registered('NMClient', '1.0');
 
-const Config = imports.misc.config;
-const Clutter = imports.gi.Clutter;
-const GLib = imports.gi.GLib;
+var smDepsGtop = true;
+var smDepsNM = true;
 
-const Gio = imports.gi.Gio;
-const Lang = imports.lang;
-const Shell = imports.gi.Shell;
-const St = imports.gi.St;
-const Power = imports.ui.status.power;
+var Config = imports.misc.config;
+var Clutter = imports.gi.Clutter;
+var GLib = imports.gi.GLib;
+
+var Gio = imports.gi.Gio;
+var Shell = imports.gi.Shell;
+var St = imports.gi.St;
+var Power = imports.ui.status.power;
 // const System = imports.system;
-const ModalDialog = imports.ui.modalDialog;
+var ModalDialog = imports.ui.modalDialog;
 
-const ExtensionSystem = imports.ui.extensionSystem;
-const ExtensionUtils = imports.misc.extensionUtils;
+var ExtensionSystem = imports.ui.extensionSystem;
+var ExtensionUtils = imports.misc.extensionUtils;
 
-const Me = ExtensionUtils.getCurrentExtension();
-const Convenience = Me.imports.convenience;
-const Compat = Me.imports.compat;
+var Me = ExtensionUtils.getCurrentExtension();
+var Convenience = Me.imports.convenience;
+var Compat = Me.imports.compat;
 
-let Background, GTop, IconSize, Locale, MountsMonitor, NM, NetworkManager, Schema, StatusArea, Style, gc_timeout, menu_timeout;
+var Background, GTop, IconSize, Locale, MountsMonitor, NM, NetworkManager, Schema, StatusArea, Style, gc_timeout, menu_timeout;
 
 try {
     GTop = imports.gi.GTop;
@@ -168,25 +168,25 @@ function interesting_mountpoint(mount) {
 }
 
 
-const smStyleManager = new Lang.Class({
-    Name: 'SystemMonitor.smStyleManager',
-    _extension: '',
-    _iconsize: 1,
-    _diskunits: _('MiB/s'),
-    _netunits_kbytes: _('KiB/s'),
-    _netunits_mbytes: _('MiB/s'),
-    _netunits_kbits: _('kbit/s'),
-    _netunits_mbits: _('Mbit/s'),
-    _pie_width: 300,
-    _pie_height: 300,
-    _pie_fontsize: 14,
-    _bar_width: 300,
-    _bar_height: 150,
-    _bar_fontsize: 14,
-    _text_scaling: 1,
+const smStyleManager = class SystemMonitor_smStyleManager {
 
-    _init: function () {
+    constructor () {
+        this._extension = '';
+        this._iconsize = 1;
+        this._diskunits = _('MiB/s');
+        this._netunits_kbytes = _('KiB/s');
+        this._netunits_mbytes = _('MiB/s');
+        this._netunits_kbits = _('kbit/s');
+        this._netunits_mbits = _('Mbit/s');
+        this._pie_width = 300;
+        this._pie_height = 300;
+        this._pie_fontsize = 14;
+        this._bar_width = 300;
+        this._bar_height = 150;
+        this._bar_fontsize = 14;
+        this._text_scaling = 1;
         this._compact = Schema.get_boolean('compact-display');
+
         if (this._compact) {
             this._extension = '-compact';
             this._iconsize = 3 / 5;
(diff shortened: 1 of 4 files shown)


shemgp:
Fixes #505 too.

p91paul:
you should probably edit metadata.json to remove all older versions then...

sibradzic:
@shemgp works here, thanks indeed! The logs complain about two instances of `array.toString() on a Uint8Array instance` conversion, but that ain't a major issue now. We need this merged...

averissimo:
great job @shemgp

kubrickfr:
With this PR applied, gnome-shell 3.32 uses 50% CPU all the time

ixevix:
Nice work but unusable because it eats up 50% CPU all the time

zegkljan:
Regarding the CPU usage, tray icon extensions (e.g. Topicons, Topicons plus, Tray icons...) have the same issue. However, one of them - Topicons redux - has the same issue but with a weird behavior where opening it's settings from gnome tweaks an odd number of times causes the CPU load to drop to normal levels (see https://gitlab.com/pop-planet/TopIcons-Redux/issues/12). What's even weirder is that this also cuts the CPU load due to system-monitor.

Apparently, Topicons redux's settings do _something_ that prevent the high CPU load.

kubrickfr:
Yes, true, I can even get to run this extension (system-monitor) without having 50% cpu usage if I play with settings and which extension is enabled enough times...

kubrickfr:
Reported upstream https://gitlab.gnome.org/GNOME/gnome-shell/issues/1054

dino99:
With the 3 above commits applied, that extension is mostly working, but:
- get this flooding the log:
JS ERROR: TypeError: this.cb is undefined
refresh/<@/usr/share/gnome-shell/extensions/system-monitor@paradoxxx.zero.gmail.com/extension.js:1496:17
- .eslintrc has been copied inside the subdir
- line 1496 is:  file.load_contents_async(null, this.cb.bind(this));
- hopefully i am not seeing high ram usage (ubuntu disco, gnome on xorg session)

sibradzic:
For some reason I don't see any unusual CPU or RAM usage patterns with this change. I have nothing but system-monitor enabled. Should not make a difference, its on Ryzen 2500U.

dino99:
Also missing schema for 'move-clock' : clock is moved to far right corner (only 1/3 visible)

shemgp:
Hopefully fixed `cb` problem.

And I didn't copy/move the `.eslintrc` to a subfolder. 
Screenshot from 2019-03-17 21-42-45

Move clock works as expected here:
* without move clock enabled
Screenshot from 2019-03-17 21-38-04
* with move clock enabled
Screenshot from 2019-03-17 21-38-24

Perhaps your icons are too plenty and fills the whole right panel so that the clock is squished?

ixevix:
Nice, seems the latest changes fixed the cpu usage problem!

dino99:
How the previous clock issue has happened:
- firstly the clock was centered 
- then i have replaced the proposed upgrade files (prefs.js, extensions.js, metadata.json) with the ones from  github (found here from the commits list)
- then created the file .eslintrc (copy/paste from here) to the extension dir (no applied chown or else)
- restarted the gnome session, and the clock  has moved to the far right (too far), and dconf warns me about a missing schema for 'move clock) 

Is that clock problem meaning that copy/paste files are not complete compared to the original ones ?

ixevix:
Ok, I spoke too soon, but disabling TopIcons solved it.

alephnull:
Confirm that this is working with gnome-shell 3.32.0 (Ubuntu Disco Dingo)

dino99:
Error logged after the 5 commits applied:

JS ERROR: TypeError: this is null
                                                cb@/usr/share/gnome-shell/extensions/system-monitor@paradoxxx.zero.gmail.com/extension.js:1492:29

shemgp:
Hopefully fixed that error around that cb function.  I don't get any errors/logs in my logs now, so please check.

dino99:
Applied the 2 commits above, and restarted the session : then glancing at journalctl. All the previous related warnings/errors are gone now. 
Good job, and thanks for the quick fixes.
Declare ready for upstream publication.

treba123:
The CPU issue has been confirmed to be an upstream issue: https://gitlab.gnome.org/GNOME/gnome-shell/issues/1054#note_462698

dino99:
Actual output logged is:

 0 cb() ["/usr/share/gnome-shell/extensions/system-monitor@paradoxxx.zero.gmail.com/extension.js":1493]

Nothing else logged. Is that '0' cb() the expected value ?

system-monitor@paradoxxx.zero.gmail.com/extension.js
```diff
@@ -635,52 +627,47 @@ const Pie = new Lang.Class({
...
-    },
-    _boxGetPreferredHeight: function (actor, forWidth, alloc) {
-        [alloc.min_size, alloc.natural_size] = this.box.get_preferred_height(forWidth);
-    },
-    _boxAllocate: function (actor, box, flags) {
-        this.box.allocate(box, flags);
-    },
-    _shift: function () {
+    }
+    // _boxGetPreferredWidth (actor, forHeight, alloc) {
```
DreaminginCodeZH: Just wondering if it is correct to just comment these methods out... [The MR to deprecate Shell.GenericContainer](https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/153) seems to suggest we can do vfunc overriding now, and [this is what I tried](https://github.com/DreaminginCodeZH/gnome-shell-system-monitor-applet/commit/19467c3e464d6b7f75989e9b1d1087e91adddbd1#diff-3d6a322a2008ae119622ecc9bde5555cR674) according to the changes in it. I didn't know if this actually preserved something, but everything also worked fine in my fork. Just hope this might be of some help :)

ivnilv:
What's holding this from being merged into master ??

M5oul:
Works fine on my side!

xkill:
Works fine on mi side too

sibradzic:
Seriously, Ubuntu 19.04 is about top be released, and so are bunch of other distros that will be using gnome shell 3.32. This needs a merge.

chrisspen:
If you just want 3.32 support, someone should submit a PR that just bumps up the version number. I'm hesitant to approve this because it makes ton of other changes that will likely break the extension for everyone else. Many people only use the LTS releases, so there's no rush to support 19.04, especially if it's going to add instability for everyone not on 19.04.

p91paul:
This pr actually drops compatibility for older gs versions. Which means that on e.g.o. users of old versions of gs will just use the old version of the extension.
When it comes to git, this should be handled with branches.
Unless you believe it is possible to support both pre and post gs 3.32 with the same codebase.

chrisspen:
Then is sounds like this PR is dead on arrival, because we're not going to drop support for everyone on distros that will be used for several more years. And maintaining a separate branch for each Gnome release is likely not something I or any other maintainer has the time for. Does Gnome 3.32 really require dropping support for all other previous versions?

matttbe:
I guess that if you choose to stay on a stable releases (e.g. Ubuntu LTS, Debian Stable or CentOS), you also want to keep everything stable, including plugins. Meaning that you also accept not to have new features, only bug fixes.

Many plugins -- including the ["official" ones](https://gitlab.gnome.org/GNOME/gnome-shell-extensions) -- add new features and small bug-fixes only for the recent versions of GNOME, simplifying also the code by not adding pieces of code working with multiple versions.

That's a bit sad plugins devs have to do some many changes to be fully compliant with the last stable release.

sibradzic:
Firs of all, I would like to express my humble respect for the dev & maintainers, I've been using this extension forever, can't imagine my Linux desktop experience without it, BIG THANK YOU PPL!

@chrisspen sais:
> If you just want 3.32 support, someone should submit a PR that just bumps up the version number.

If it was as simple as bumping up version, you think @shemgp would be going through all this re-factoring just for the sake of the thrill? These changes are *a must* for gnome-shell >=3.32, but they don't break compatibility with all previous versions, please read further...

@p91paul sais:
> Then is sounds like this PR is dead on arrival, because we're not going to drop support for everyone on distros that will be used for several more years. And maintaining a separate branch for each Gnome release is likely not something I or any other maintainer has the time for. Does Gnome 3.32 really require dropping support for all other previous versions?

These changes are not really breaking compatibility with pre-3.32 gnome-shell. The changes fix long-standing issue of hacks that were necessary before ES6 (aka ECMAScript 6 - ECMAScript 2015) implementation was supported by gnome, which landed *before* 3.32. What 3.32 does is deprecation of these hacks. I bothered to double-check the compatibility myself, @shemgp 's master branch (
https://github.com/shemgp/gnome-shell-system-monitor-applet, with the changes in question) *works perfectly fine* on Ubuntu 18.04 / gnome-shell  3.28.

Now, according to https://ptomato.wordpress.com/2017/07/14/inventing-gobject-es6-classes/, the compatibility may be broken only for gnome-shell older than 3.26 (someone can test?). This means that only "ancient" gnome desktops may be affected.

A friendly suggestion here is to branch the old release for "ancient" gnome-shell support (<3.26), and merge @shemgp 's changes to master. This is the only reasonable action really, the support for hackish GObject classes will never come back to gnome-shell, most other extensions already updated accordingly, so is the time for gnome-shell-system-monitor-applet.

system-monitor@paradoxxx.zero.gmail.com/metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-    "shell-version": ["3.4","3.6","3.8","3.10", "3.12", "3.14", "3.16", "3.18", "3.20", "3.22", "3.24", "3.26", "3.28", "3.30"],
+    "shell-version": ["3.32"],
```
sibradzic: Looks like ES6 is supported starting from gnome-shell 3.26. I've tested your changes in Ubuntu 18.04 / gnome-shell 3.28 and everything works. I guess this should be:
`"shell-version": ["3.26", "3.28", "3.30", "3.32"],`
instead?

chrisspen:
@sibradzic Thanks for the clarification. You bring up some very good points.

PR #512, opened 2019-03-17
Fix #504  (array.to string() warnings)

### LINKED ISSUE #504: array.toString() on a Uint8Array
On Manjaro 18.0.2 with GNOME Shell 3.30.2, journalctl gives the following message:

> Some code called array.toString() on a Uint8Array instance. Previously this would have interpreted the bytes of the array as a string, but that is nonstandard. In the future this will return the bytes as comma-separated digits. For the ti...
> (Note that array.toString() may have been called implicitly.)
> 0 refresh/<() ["/home/steve/.local/share/gnome-shell/extensions/system-monitor@paradoxxx.zero.gmail.com/extension.js":2066]
> 1 () ["self-hosted":979]

Immediately prior to this there is an entry:

> Some code accessed the property 'UPower' on the module 'power'. That property was defined with 'let' or 'const' inside the module. This was previously supported, but is not correct according to the ES6 standard. Any symbols to be exported ...

The monitor itself seems to be working, but I couldn't see any other mention of this, so thought I would raise a ticket.

rschupp:


chrisspen:
Could you please resolve conflicts?

PR #513, opened 2019-03-18
Add gnome-shell install check to travis-ci config.

chrisspen:
I've extended our Travis-CI configuration to do a test installation of the extension into Gnome-Shell as well as ensure its CPU and memory usage doesn't exceed certain limits.


chrisspen:
Any objections to merging this?

PR #516, opened 2019-04-28
Fedora

NVieville:
Hello,

These two commits were necessary to build Fedora package (fc30). 

The first one fixes a typo in Makefile DESTDIR variable when used for admin users.

The second one introduces one new variable BUILD_FOR_RPM in Makefile to not use 'sudo' and 'gnome-shell-extension-tool' only for this case. This variable have to be set to '1' for both build and install target of the Makefile to build a RPM package.

Comment about this are welcome.

Cordially,

-- 
NVieville


chrisspen:
Thanks for the update.

PR #518, opened 2019-05-01
remove obsolete compatibility code...

rschupp:
...that refers to versions older than the versions currently in
"shell-version" from metadata.json

Note: The old "if (Compat.versionCompare(shell_Version, 'x.y.z')) ..."
stuff never really worked as intended: Compat.versionCompare()
(from compat.js) returns -1, 0 or 1 (or NaN) and *not* a boolean.


chrisspen:
Thanks.

PR #544, opened 2019-11-04
Fix typo nvidia-settings in gpu_usage.sh

NVieville:
Signed-off-by: NVieville <nicolas.vieville@uphf.fr>

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/gpu_usage.sh
+++ b/system-monitor@paradoxxx.zero.gmail.com/gpu_usage.sh
@@ -31,7 +31,7 @@ checkcommand()
 if checkcommand nvidia-smi; then
 	nvidia-smi -i -1 -q -d MEMORY | grep -A4 -i gpu | egrep -i "used|total" | awk '{print $3}'
 	# This line will print the GPU usage in %.
-	sed -e 's#.*=\(\)#\1#;s'/.$//'' <<< $(nvidia-set//tings -q all | grep -i GPUUtilization | grep graphics | awk '{print $4}')
+	sed -e 's#.*=\(\)#\1#;s'/.$//'' <<< $(nvidia-settings -q all | grep -i GPUUtilization | grep graphics | awk '{print $4}')
 elif checkcommand glxinfo; then
 	TOTALVRAM="`glxinfo | grep -A2 -i GL_NVX_gpu_memory_info | egrep -i "dedicated" | cut -f2- -d ':' | gawk '{print $1}'`"
 	AVAILVRAM="`glxinfo | grep -A4 -i GL_NVX_gpu_memory_info | egrep -i "available dedicated" | cut -f2- -d ':' | gawk '{print $1}'`"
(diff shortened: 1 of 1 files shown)


NVieville:
Hello,

While this pull request fixes a typo, I have another fix which I am not sure is necessary for all and will work with all versions of nvidia-smi. 
The line 32 of the gpu_usage.sh file doesn't give satisfaction on this configuration (tooltip showing NaN regularly):
- Fedora 31 x86_64
- GF100GL [Quadro 4000]
- NVidia driver and tools 390.129

Modifying it to:

`nvidia-smi -q -d MEMORY | grep -A4 -i gpu | egrep -i "used|total" | awk '{print $3}'`

does. It's also possible to get the same result to modify it to:

`nvidia-smi --query-gpu=memory.total,memory.used --format=csv,noheader,nounits | sed 's%, %\n%g'`

I can't say if the last one is specific to one version of nvidia-smi tool, and if it will work in any cases.
These two lines doesn't address a specific device, and can with some hardware configurations give results not usable in the extension. Maybe "-i 0" should be inserted in each line in order to be sure to give usable results, but just for the first detected NVidia device. 
In this case, these lines would look like this:

`nvidia-smi -i 0 -q -d MEMORY | grep -A4 -i gpu | egrep -i "used|total" | awk '{print $3}'`

and

`nvidia-smi -i 0 --query-gpu=memory.total,memory.used --format=csv,noheader,nounits | sed 's%, %\n%g'`

Please, feel free to make any comment about this subject.

Cordially,


-- 
NVieville

NVieville:
Hello,

Sorry for the noise, but is there any news about this issue.
We're hitting our first bug report in Fedora:

https://bugzilla.redhat.com/show_bug.cgi?id=1794158

Feel free to make any comment about this subject.

Cordially,


-- 
NVieville

encounter:
I'm also hitting the NaN issue with a 2070 SUPER and NVIDIA driver 440.44.

Removing `-i 0` as suggested by NVieville does indeed fix it.

encounter:
A further improvement might be 
```
nvidia-smi -i 0 --query-gpu=memory.total,memory.used,utilization.gpu --format=csv,noheader,nounits | sed 's%, %\n%g'
```
which prints total memory, used memory, and GPU utilization all in one fell swoop (removes the need for the nvidia-settings call)

NVieville:
Hello,

Sorry again for the noise, but is there any news about this issue. Do you think this pull request will be merge in the near future?

Cordially,

--
NVieville

RenWal:
Confirmed this bug on a Dell G5 15 5590 running dual Intel / NVidia RTX graphics on Ubuntu 20.04.

The change from this PR works just fine for me and is WAY faster than the current implementation. This is a good thing since I had issues with the GPU just freezing momentarily while an `nvidia-settings` query is running, causing frame drops every time the GPU usage is queried.

franglais125:
Can somebody confirm whether this fix is fine on older models/drivers? i.e. backwards compatible.

I'm the on who introduced this feature, but I never owned an Nvidia GPU, so I rely entirely on everyone's feedback.

RenWal:
I can't say anything about old nvidia drivers, but I made this change on my other machine running Ubuntu 18.04.4 and `nvidia-driver-430` package containing the driver 430.64 from `bionic-updates restricted` with a GTX 1060 and it ran reliably for about a day now.

Issue #496, opened 2019-01-25
synchronous IO (read) blocks entire GNOME-Shell (desktop)


return42:
In extensions calling synchronous IO functions like `Shell.get_file_contents_utf8_sync` will block  GNOME-Shell's main loop. This means nearly everything is blocked: mouse feels sticky, videos are stutter, keyboard-repeat lags and so on.

I can not say for sure, but I suspect issues (like) #433, #439, #417, #350, #368, #263, #235, #207, #202 and #187 are related to such synchronous IO.

I will try to get rid of these synchronous calls in my fork and come back.

return42:
@chrisspen thanks for merging PR #497 ! But this was only the first step, there are some sync-IO calls left (especially Disk.refresh). As soon as I have time for this, I want to take care of it. So my question is, can we left this issue open for the first?  / Thanks!

chrisspen:
Sure, Github auto-closed the referenced PR when merged. I'll reopen.

pwaller:
@return42 - I just found myself here. Suddenly noticed my buttery smooth system was not behaving well, and happened to realise that what I did recently was enable the system monitor. Just did some investigating and found my way to #202 and this PR. So glad to see someone tackling it, the pauses are pretty horrendous on my powerful system, causing missed inputs and all sorts of things!

Just wanted to say thanks for working on this :heart:

pwaller:
By the way, I just tried out the master branch which has #497 merged. The Jankiness is gone :rocket: , but now the CPU freq monitor doesn't work. It's just empty. If I show digits, there is nothing shown.

return42:
@pwaller sorry for delay - I just pulled master from origin but can't reproduce any issue with CPU freq. I'am sorry for not having any clue whats going wrong here. What the extension does to get CPU freq is very simple; it reads the vales from `/sys/`. Can you please check if you get some CPU freq values (kHz):

     $ cat /sys/devices/system/cpu/cpu*/cpufreq/scaling_cur_freq
     1374140
     1361529
     ...

And can you please check your GNOME shell version, here I tested with:

    $ gnome-shell --version
    GNOME Shell 3.28.3

pwaller:
Yes, that does show frequencies. I'm get the same version number back.

return42:
hm, I do not have any clue .. very sorry  .. I can only ask common places like; what dist you are using, can you reproduce this issue with another system and/or user login?

return42:
[...]

PR #545, opened 2019-11-04
Get rid of synchronous Disk read

datumbox:
I've been experiencing freezes while using this extension and after reading the issues #496 and #207 I understood that they might be related to the synchronous Disk reads that can block the entire GNOME-shell. In this PR Ι follow a similar approach as @return42 to make the disk reads asynchronous. I tested it on my PC and it seems to work fine.

Please note that I don't typically code in JavaScript and I've never developed a Gnome Extension, so you can safely assume I have no idea what I'm doing here. **Please review carefully.**

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -1381,27 +1381,33 @@ const Disk = class SystemMonitor_Disk extends ElementBase {
     }
     refresh() {
         let accum = [0, 0];
-        let lines = Shell.get_file_contents_utf8_sync('/proc/diskstats').split('\n');
 
-        for (let i = 0; i < lines.length; i++) {
-            let line = lines[i];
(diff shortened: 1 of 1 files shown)


ioogithub:
I am also experiencing system freezes and I have been trying to track it down all week. My mouse, keyboard, everything freezes from a fraction of a second to up to 10 seconds. I have used this extension on Fedora 30, Fedora 31 and Debian 10.1 and 10.2. I have seen freezing on all of them however it is much worse on Debian 10.2.

I also found this mailing list discussion: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=889706

I suspect it is this extension. I have it off now and have not had a freeze yet.

datumbox:
@chrisspen I'm not sure if you can review this one. Let me know if you think we can merge this to master. I've been using the patch since November and it runs smoothly.

franglais125:
Merged as [0386578905decea1106425e8a6ba7ab0ace9bb2c]

PR #558, opened 2020-01-08
Update installation instructions for Ubuntu

mominul:
Package `gir1.2-networkmanager-1.0` is not available.

Screenshot from 2020-01-09 00-51-07

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Before installing this extension, ensure you have the necessary system packages 
 
 * On Ubuntu:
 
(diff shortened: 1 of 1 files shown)


mominul:
Distro: Ubuntu 19.10

Saumen95:
Look at the 2nd last message.try it.Let me know if this trick solves your problem

mominul:
@Saumen95 Yes, it works with `gir1.2-nm-1.0` which I have included in my patch.

Issue #567, opened 2020-03-13
Doesn't work with GNOME Shell 3.36.0


nemoload:
Mar 13 16:26:55 eva-00 gnome-shell[7226]: JS ERROR: TypeError: this.tip_labels[i] is undefined
                                          update@/home/ahmed/.local/share/gnome-shell/extensions/system-monitor@paradoxxx.zero.gmail.com/extensions/system-monitor@paradoxxx.zero.gmail.com/extension.js:925:56

mcardillo55:
Looks like this is coming from the following exception:
JS ERROR: Extension system-monitor@paradoxxx.zero.gmail.com: Error: Tried to construct an object without a GType; are you using GObject.registerClass() when inheriting from a GObject type?

It looks like it's stemming from checks that were introduced by gjs. There is more discussion and proposed fixes in the thread and comment linked: https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/issues/551#issuecomment-558302621

mcardillo55:
Opened a PR. Works for me on 3.36. Would appreciate feedback on 3.36 and previous versions.

mcardillo55:
Sorry, did not realize PR #564 was already open. That one works for me as well.

Kerrung:
Still doesn't work with GNOME Shell 3.36.0 on Arch Linux... When it be changed?

nemoload:
@Kerrung It's not merged it but you can build it yourself from PR #564

Kerrung:
> @Kerrung It's not merged it but you can build it yourself from PR #564

But, when it will be merged?

leinardi:
Is this project abandoned? Last commit was on November 2019...

will-ca:
@leinardi It's still quite alive AFAICT. The core developers seem to be taking their time to prioritise stability and quality over pushing out a release ASAP.

will-ca:
@leinardi It's still quite alive AFAICT. The core developers seem to be taking their time to prioritise stability and quality over pushing out a release ASAP.

leinardi:
@will-ca prioritize stability and quality is not against the release of an hotfix for a critical bug that make the plugin crash on every recent Linux distribution.

ciltocruz:
Hello.
I have updated to ubuntu 20.04 with GNOME 3.36.1 and I get the same error.

The extension does not load.

[...]

PR #564, opened 2020-03-03
Make it work with gnome-shell 3.36

shemgp:
Well actually it's for GNOME Shell 3.35.91 still, but hopefully no more changes required for 3.36.

References for where I got hints for the fixes:
* `TipItem` error (Tried to construct an object without a GType): http://gjs.guide/guides/gjs/transition.html#extending-gobject-classes
* removal of `raise_top`: https://github.com/paperwm/PaperWM/issues/212
* Depreciated `TableLayout`: https://mail.gnome.org/archives/commits-list/2014-August/msg01899.html

PR #568, opened 2020-03-19
Updates for Gnome 3.36

mcardillo55:
Addresses errors in #567. Tested on 3.36, would appreciate feedback for previous versions to see if metadata.json should be corrected.


mcardillo55:
Did not notice #564 which seems to add a few other changes as well. Closing.

PR #569, opened 2020-03-28
Fix GPU display.

peci1:


powerpaul17:
Works for me

jonnyeom:
Thanks, applied the patch against the latest 3.36 PR and it works great!

franglais125:
Fixed in #544.

PR #578, opened 2020-04-13
Prevent refreshes while the Shell is busy.

franglais125:
We pass GLib.PRIORITY_DEFAULT_IDLE as priority to Mainloop to avoid
micro-freezes of the Shell. This is mostly visible in Wayland sessions.


DevDef:
@chrisspen I’ve just tried this. I don’t know if it actually works, but it does no harm. So I’m in favour of merging this PR.

PR #584, opened 2020-04-25
sanitize usage vals parsed from gpu_usage.sh script

prehensile:
avoids NaNs being displayed in widget when script returns unexpected values (e.g GPU is powered down etc)


DevDef:
@chrisspen This seems a pretty good idea. Please merge the PR.

PR #586, opened 2020-04-26
gpu_usage.sh: Remove superfluous function return statement

d97jro:
From bash's man page: "the exit status of a function is the exit
status of the last command executed in the body".


DevDef:
It makes sense. @chrisspen please merge this.

PR #587, opened 2020-04-26
gpu_usage.sh: Use contemporary $() instead of legacy ``

d97jro:
It is recommended to use $() over the legacy ``.
See e.g.
https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xcu_chap02.html#tag_23_02_06_03
Also removed the quotes that are not needed when assigning variables.

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/gpu_usage.sh
+++ b/system-monitor@paradoxxx.zero.gmail.com/gpu_usage.sh
@@ -31,8 +31,8 @@ checkcommand()
 if checkcommand nvidia-smi; then
 	nvidia-smi -i 0 --query-gpu=memory.total,memory.used,utilization.gpu --format=csv,noheader,nounits | sed 's%, %\n%g'
 elif checkcommand glxinfo; then
-	TOTALVRAM="`glxinfo | grep -A2 -i GL_NVX_gpu_memory_info | grep -E -i "dedicated" | cut -f2- -d ':' | gawk '{print $1}'`"
-	AVAILVRAM="`glxinfo | grep -A4 -i GL_NVX_gpu_memory_info | grep -E -i "available dedicated" | cut -f2- -d ':' | gawk '{print $1}'`"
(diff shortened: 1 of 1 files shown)


DevDef:
I don’t see the real benefit from this change but, anyway, @chrisspen could you merge it please?

chrisspen:
In general, I reject any changes that are functionally equivalent to the current behavior, yet provide no clear benefit. We shouldn't change the code for change's sake. It should make the code more efficient or fix functionality on a wider set of platforms.

What is the benefit of a more "contemporary" syntax? Is the current syntax not supported anymore or deprecated?

d97jro:
The reason to use `$()` over backticks (which is ancient and since long recommended not to use) is very well described here:

https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xcu_chap02.html#tag_23_02_06_03
https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_03

chrisspen:
I'm aware the feature is old, and I understand that some people dislike the syntax, but your links don't answer my question. How does this change fix a current problem?

d97jro:
Oh, normally when people use backticks they are not aware of the `$()` syntax and why it is recommended to use it instead (for the reasons described in those links). They tend to get the backticks from copying some other script that copied it from some other script that got it from someone who found some old snippet from Stackoverflow or such. If it is a deliberate active choice to have backticks then it is of course a different situation, although I have yet to encounter someone who _actively choose_ backticks when they are aware of `$()`.

No, this of course does not fix any severe problems. Were you to add tools like shellcheck or such to the CI pipeline, you would have to add exceptions for allowing the backticks syntax, but currently, no, this is solely a refinement PR.

If you do not want this, please feel free to close the PR.

chrisspen:
Ok, thanks. That's what I was looking for. If this makes the code a little more future-proof, then that's an improvement.

PR #588, opened 2020-04-26
gpu_usage.sh: Use faster shell internals instead of sed

d97jro:
The sed command that changes ', ' from the `nvidia-smi` output with
newlines can be replaced with shell built-in functionality.

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/gpu_usage.sh
+++ b/system-monitor@paradoxxx.zero.gmail.com/gpu_usage.sh
@@ -28,7 +28,7 @@ checkcommand()
 # This will print three lines. The first one is the the total vRAM available,
 # the second one is the used vRAM and the third on is the GPU usage in %.
 if checkcommand nvidia-smi; then
-	nvidia-smi -i 0 --query-gpu=memory.total,memory.used,utilization.gpu --format=csv,noheader,nounits | sed 's%, %\n%g'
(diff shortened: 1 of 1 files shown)


psukys:
is performance that much of an issue for a single line parsing?

d97jro:
> is performance that much of an issue for a single line parsing?

The suggested is 2x faster (or uses 50% the CPU, depending on how you see it).

It is possible to use one line only with

`while IFS=', ' read -r a b c; do echo "$a\n$b\n$c"; done`

instead, would you prefer that?

DevDef:
@d97jro I personally would prefer a oneliner to fit the indentation.
But I think it should rather be:
` IFS=', ' read -r a b c; do echo -e "$a\n$b\n$c"; done`
Could you test it and commit the change so that it appears in this PR?

d97jro:
The `echo -e` is not POSIX and will potentially cause issues. Then I reckon using a `printf` would be the safer choice for a oneliner:

`IFS=', ' read -r a b c; do printf '%s\n%s\n%s\n' "$a" "$b" "$c"; done`

d97jro:
> is performance that much of an issue for a single line parsing?

For a measuring/monitoring tool, yes absolutely, it should use as few clock cycles as possible.
Especially for tasks that are frequently run.

d97jro:
> `IFS=', ' read -r a b c; do printf '%s\n%s\n%s\n' "$a" "$b" "$c"; done`

...or perhaps

`IFS=', ' read -r a b c; do echo "$a"; echo "$b"; echo "$c"; done`

DevDef (approved):
@chrisspen Please, merge it.

PR #589, opened 2020-04-26
gpu_usage.sh: Use (( expr )) instead of let expr

d97jro:
Use (( expr )) instead of let expr


DevDef:
Why?

d97jro:
This PR is superfluous since https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/pull/591 exists.

PR #590, opened 2020-04-26
gpu_usage.sh: Used grep -E instead of egrep

joakimr-axis:
egrep is considered non-standard and deprecated, and it is recommended
to use grep -E instead.


DevDef:
@chrisspen This makes sense and is safe. Please, merge this PR.

PR #591, opened 2020-04-26
gpu_usage.sh: Make POSIX compatible and use /bin/sh

d97jro:
Using /bin/sh instead of bash will make this script run much more
lightweight on hosts with e.g. Debian-based systems that uses dash for
/bin/sh which is ~10% the footprint of bash.


DevDef:
@chrisspen tiny improvement, but let’s merge it.

DevDef:
@chrisspen Could you merge this PR please?

PR #593, opened 2020-04-27
GPU usage: show usage when using glxinfo output

psukys:
Previously glxinfo only output total and used memory when using glxinfo, but the nvidia output as well contained usage information, this pull request simply adds that additional third - utilization - data outptu

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/gpu_usage.sh
+++ b/system-monitor@paradoxxx.zero.gmail.com/gpu_usage.sh
@@ -36,4 +36,5 @@ elif checkcommand glxinfo; then
 	let FREEVRAM=TOTALVRAM-AVAILVRAM
 	echo "$TOTALVRAM"
 	echo "$FREEVRAM"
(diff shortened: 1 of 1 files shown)


DevDef:
Your calculation relies on `bc`. This is adding a dependency. Could you make it without bc?

DevDef:
The following is POSIX shell without dependencies (although it’s limited to integer values):
`echo $(( FREEVRAM * 100 / TOTALVRAM ))`

psukys:
bc is part of posix

DevDef:
It’s not installed by default on all systems, so it is a problem. As managing external dependencies on each and every system is painful, it’s better not to add one more dependency that is not essential.

chrisspen:
For the reason @DevDef has pointed out, I don't think change is appropriate. You're introducing an external tool to do something as trivial as calculate a decimal. Javascript can easily do this.

If you want to calculate this in Javascript or a more universally supported method, that wouldn't be a problem. Closing this for now. Feel free to re-open if you want to update it.

PR #598, opened 2020-04-28
gpu_usage.sh: Replace cut and gawk with shell internals

d97jro:
The use of cut and gawk in the script can easily be replaced with basic
POSIX shell script internal parameter expansion functionality. The
benefit of this is portability, since there are no longer dependencies
on cut and gawk. (And as a bonus, execution of the script gets more
lightweight.)

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/gpu_usage.sh
+++ b/system-monitor@paradoxxx.zero.gmail.com/gpu_usage.sh
@@ -30,9 +30,13 @@ checkcommand()
 if checkcommand nvidia-smi; then
 	nvidia-smi -i 0 --query-gpu=memory.total,memory.used,utilization.gpu --format=csv,noheader,nounits | while IFS=', ' read -r a b c; do echo "$a"; echo "$b"; echo "$c"; done
 elif checkcommand glxinfo; then
-	TOTALVRAM=$(glxinfo | grep -A2 -i GL_NVX_gpu_memory_info | grep -E -i "dedicated" | cut -f2- -d ':' | gawk '{print $1}')
-	AVAILVRAM=$(glxinfo | grep -A4 -i GL_NVX_gpu_memory_info | grep -E -i "available dedicated" | cut -f2- -d ':' | gawk '{print $1}')
-	FREEVRAM=$((TOTALVRAM-AVAILVRAM))
+	TOTALVRAM=$(glxinfo | grep -A2 -i GL_NVX_gpu_memory_info | grep -E -i 'dedicated')
+	TOTALVRAM=${TOTALVRAM##*:[[:blank:]]}
+	TOTALVRAM=${TOTALVRAM%%[[:blank:]]MB*}
+	AVAILVRAM=$(glxinfo | grep -A4 -i GL_NVX_gpu_memory_info | grep -E -i 'available dedicated')
+	AVAILVRAM=${AVAILVRAM##*:[[:blank:]]}
+	AVAILVRAM=${AVAILVRAM%%[[:blank:]]MB*}
+	let FREEVRAM=TOTALVRAM-AVAILVRAM
 	echo "$TOTALVRAM"
 	echo "$FREEVRAM"
 fi


system-monitor@paradoxxx.zero.gmail.com/gpu_usage.sh
```diff
@@ -30,8 +30,12 @@ checkcommand()
 if checkcommand nvidia-smi; then
 	nvidia-smi -i 0 --query-gpu=memory.total,memory.used,utilization.gpu --format=csv,noheader,nounits | sed 's%, %\n%g'
 elif checkcommand glxinfo; then
-	TOTALVRAM=$(glxinfo | grep -A2 -i GL_NVX_gpu_memory_info | grep -E -i "dedicated" | cut -f2- -d ':' | gawk '{print $1}')
-	AVAILVRAM=$(glxinfo | grep -A4 -i GL_NVX_gpu_memory_info | grep -E -i "available dedicated" | cut -f2- -d ':' | gawk '{print $1}')
+	TOTALVRAM=$(glxinfo | grep -A2 -i GL_NVX_gpu_memory_info | grep -E -i "dedicated")
+	TOTALVRAM=${TOTALVRAM##*:[[:blank:]]}
+	TOTALVRAM=${TOTALVRAM%%[[:blank:]]MB*}
```
DevDef, line 35: These 2 lines could be replaced by one:
`TOTALVRAM=${TOTALVRAM//[^[:digit:]]/}`

system-monitor@paradoxxx.zero.gmail.com/gpu_usage.sh
```diff
@@ -30,8 +30,12 @@ checkcommand()
...
+	AVAILVRAM=$(glxinfo | grep -A4 -i GL_NVX_gpu_memory_info | grep -E -i "available dedicated")
+	AVAILVRAM=${AVAILVRAM##*:[[:blank:]]}
+	AVAILVRAM=${AVAILVRAM%%[[:blank:]]MB*}
```
DevDef, line 38: These 2 lines could also be replaced by one:
`AVAILVRAM=${AVAILVRAM//[^[:digit:]]/}`

DevDef (changes requested):

system-monitor@paradoxxx.zero.gmail.com/gpu_usage.sh
d97jro: Yes, but that is bash-specific syntax and not POSIX, which would have to be replaced with these two lines in order to get e.g. https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/pull/591 in place. And there is no need to use bash when the job can be done with 10% of the computing resources and footprint with a POSIX shell such as dash (default /bin/sh shell in Debian-based distributions since a couple of releases back).

system-monitor@paradoxxx.zero.gmail.com/gpu_usage.sh
d97jro: Please see reply to comment below.

DevDef:
So you mean `${TOTALVRAM##*:[[:blank:]]}` and `${TOTALVRAM%%[[:blank:]]MB*}` are POSIX, but `TOTALVRAM=${TOTALVRAM//[^[:digit:]]/}` is pure bashism? I’m surprised by this. BTW, do you have a document or a tool to check this POSIX compliance?

d97jro:
Yes, that substitution style is a bashism and not POSIX.

One very good source to see what is POSIX or not is `man dash` since dash is a POSIX (only) shell. That man page is very well written and simple enough to navigate around in. If you compare it to bash's man page it becomes really apparent what huge a beast bash is—sometimes one wants and needs the extras that bash brings to the table, but more often than not that is not the case.

One (there are several others) very quick and convenient way to test if your script is POSIX compliant on a system that uses bash as its `/bin/sh` is to just change the shebang to e.g. `#!/bin/dash` to explicitly tell it to run with dash. The non-POSIX substitution you suggest here will then generate the error `Bad substitution` and the script fail.

There are of course tools that will assist you in this, one being [checkbashisms](https://linux.die.net/man/1/checkbashisms) which is available in the regular Linux distributions and quite conveniently does what it says: checks for bashisms. Another is [shellcheck](https://www.shellcheck.net/) which will check your shebang and warn if you add bashisms in a script that has a `#!/bin/sh` shebang.

DevDef:
OK, thanks for the hints.
@chrisspen Eventually, I think you can merge this PR as is.

chrisspen:
Ok. I'm testing locally. If everything looks good, I'll merge in a few days.

DevDef:
@chrisspen It’s been a while now. Did you have troubles? We were in a good rhythm in the PR merging process. It’s stalled now. :-/

chrisspen:
Sorry for the long delay. Things became hectic in my day job, and this just fell through the cracks.

PR #609, opened 2020-05-13
Replace the byte array to utf-8 string conversion routine.

katox:
This patch fixes the NaN numbers in the monitored stats, like Thermal or Disk.

Reason: The original ByteArray.toString doesn't properly convert
the utf-8 encoded bytes into a string. Neither just calling the String
constructor does the proper decoding.


chrisspen:
Closing due to failing build. Feel free to re-open if you can fix the conflicts.

PR #612, opened 2020-05-14
Added the possibility to rearrange the graphs

### LINKED ISSUE #245: Order of items changed
Since Gnome 2.x days, the order was:
CPU - Memory - Net - Swap - Disk

For some reason this has now changed to:
CPU - Memory - Swap - Net - Disk

Either change this back, or make it configurable.

Thanks!

DColadas:
I added some XML attributes and changed some code in `enable()` and the Settings window, making it possible to reorder the graphs (shown in the status bar and the information window) and the settings in the Settings window. This fixes #245 and is a nice feature.

To start rearranging, go to Settings, grab the tab label you wish to move, and drag it to its position. The changes in the order of the tabs are immediate (they will remain the next time the window is opened), but the displays in the status bar will only move after restarting GNOME Shell (`Alt+F2` and the command `r`), although I may be able to work on improving it. The default order is the same as the previous one.

If you have some improvements (a better option than drag and dropping, for instance) or suggestions, feel free to comment!

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -2338,19 +2338,34 @@ function enable() {
         };
 
         // Items to Monitor
-        Main.__sm.elts = createCpus();
-        Main.__sm.elts.push(new Freq());
-        Main.__sm.elts.push(new Mem());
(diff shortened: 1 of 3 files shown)


chrisspen:
Closing due to failing build. Feel free to re-open if you can fix the conflicts.

PR #614, opened 2020-05-21
AMD GPU support via ROCm-smi

brydon:
Allow support for AMD GPUs that use ROCm drivers.


ZimbiX:
Having to launch a Python process each tick would have some CPU overhead. Does #641 work for you?

chrisspen:
Closing due to failing build. Feel free to re-open if you can get the build to pass.

PR #630, opened 2020-09-18
Add element for battery charge/discharge rate

adamsmd:
This pull request adds an element to show the current charging or discharging rate.  It is based on the contents of `/sys/class/power_supply/BAT0/`.  The following image shows an example chart with green for charging and red for discharging:

image

The settings page for this element is shown in the following image:

image

(Edit: Added description that was previously empty.)


adamsmd:
@paradoxxxzero: This pull request has been pending for a while, and I wondered if you have any plans to accept or reject it.

chrisspen:
Sorry for the delay in reviewing this. Unfortunately, the new charging section just renders "[object Gj..."

Screenshot from 2021-07-05 15-17-07

chrisspen:
I would question the need for this feature, since current Gnome already includes something very similar to this. It's not a graph, as you've implemented, but it does estimate the rate of recharge/discharge.

adamsmd:
I have not been able to reproduce the "[object Gj..." bug.  I would assume it has something to do with lines 2332-2336 (https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/pull/630/files#diff-eb3370507dafe138d1257bbbbc558a6a0872d4e7bbc818c6e2b408268a3be8d5R2332-R2336).  Is there any details you can provide?

adamsmd:
With regard to the need for this, what gnome feature are you referring to?  I see where it shows whether it is charging or discharging and the current battery level (e.g., image), but not where it shows charging rate.

chrisspen:
I'm not sure what other details I can provide. Did you test on Ubuntu 18.04? Ubuntu 20.04 switched Gnome versions that include a ton of changes, so if that's what you're developing on, you won't be able to reproduce.

And yes, you've identified the feature I'm referring to. If you click on that, it shows the charging and discharging rate. At least, it does for me in Gnome-Shell running on Ubuntu 18.04. I don't know if they've removed this or moved it around in later versions.

adamsmd:
I am currently on Ubuntu 21.04, and now that you mention it, I vaguely recall having to fix the code around when I upgraded to 20.04.  Is there a way you recommend for testing old versions (e.g., use a VM)?

On my machine, the Gnome feature you refer to shows something like "2:13 Remaining (83%)" when discharging and "0:18 Until Full (84%)" when charging.  However, the display this patch adds shows the current power usage in Watts.  The former tells me how long before I run out.  The latter tells me the rate of usage.

I find this useful to be able to monitor my machine for abnormal power usage (similar to how the CPU graph is useful for monitoring for abnormal CPU usage).  The time and percentage remaining that Gnome provides doesn't help with that in my experience.  (In theory, I could divide the percentage remaining by the percent charge remaining that Gnome provides, but that (1) would require mental math, (2) would require frequently checking to open the menu and check the values, and (3) may have confounding variables that a simple watts display does not.)

chrisspen:
If you run `test.sh`, does it pass for you locally? That runs some simple UI checks in Docker running Ubuntu 20. That's also run by the build, but first conflicts in the PR will need to be resolved.

chrisspen:
Closing as unnecessary, due to this feature now being part of Gnome. Feel free to comment if I'm mistaken.

PR #642, opened 2020-11-06
Add --yes to apt and --non-interactive to zypper

Bengt:
This increases consistency with the --assume-yes flag in the yum call.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,11 +19,11 @@ Before installing this extension, ensure you have the necessary system packages 
 
 * On Ubuntu:
 
-      $ sudo apt install gir1.2-gtop-2.0 gir1.2-nm-1.0 gir1.2-clutter-1.0
+      $ sudo apt install --yes gir1.2-gtop-2.0 gir1.2-nm-1.0 gir1.2-clutter-1.0
       
 * On Debian:
 
-      $ sudo apt install gir1.2-gtop-2.0 gir1.2-nm-1.0 gir1.2-clutter-1.0
+      $ sudo apt install --yes gir1.2-gtop-2.0 gir1.2-nm-1.0 gir1.2-clutter-1.0
 
 * On Fedora:
 
@@ -31,7 +31,7 @@ Before installing this extension, ensure you have the necessary system packages 
     
 * On openSUSE (Leap 42.1):
 
-      $ sudo zypper install gnome-shell-devel libgtop-devel libgtop-2_0-10
+      $ sudo zypper --non-interactive install gnome-shell-devel libgtop-devel libgtop-2_0-10
 
 Additionally, if you have an Nvidia graphics card, and want to monitor its memory usage, you'll need to install `nvidia-smi`.
 
--- a/README.md
+++ b/README.md
@@ -19,19 +19,19 @@ Before installing this extension, ensure you have the necessary system packages 
 
 * On Ubuntu:
 
-      $ sudo apt install --yes gir1.2-gtop-2.0 gir1.2-nm-1.0 gir1.2-clutter-1.0
+      $ sudo apt install gir1.2-gtop-2.0 gir1.2-nm-1.0 gir1.2-clutter-1.0
       
 * On Debian:
 
-      $ sudo apt install --yes gir1.2-gtop-2.0 gir1.2-nm-1.0 gir1.2-clutter-1.0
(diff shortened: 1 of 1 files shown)


mgalgs:
It's incredibly rare to see `--yes` (and similar) in installation instructions. Many users (myself included) prefer to preview the package manager transaction for new software before letting 'er fly. For automated scripts and build systems the non-interactive options are great, of course.

Regarding:

> This increases consistency with the --assume-yes flag in the yum call.

you do make a good point. However, it's interesting to note that the commit that originally introduced the `--assume-yes` flag (merged in #243) was already inconsistent in this regard. My guess is that the reviewer of that PR just didn't notice it. Had the original author been consistent and had some kind of `--yes` flag for each distro it may have stuck out more and possibly not been merged like that.

I'm just guessing what was in other peoples' heads at this point though (what else do I have to do with my time on a lovely Friday afternoon? :joy: )... But in my humble opinion a patch _removing_ the `--assumeyes` would be more useful.

mgalgs:
lmao just noticed you were the author of that original patch :rofl: :rofl: Long time contributor, that's awesome. I love this extension.

Anyhow, I still think removing the `--assumeyes` would be better since it would increase consistency with the wider ecosystem.

Bengt:
> lmao just noticed you were the author of that original patch rofl rofl Long time contributor, that's awesome. I love this extension, thank you for your work.

Ahaha, I had a a feeling like I have done this before. However, I could not quite put a finger on it. I thought, maybe I had filed a pull request like this before. Only one way to find truth on the internet: Post untruth. ;) - You are very welcome and thank you weighing in on this. Judging from the committer page, there is a very long tail of 1- 2- and 3-commit authors in this project. I am glad to be part of this community effort.

[My commit in question for reference](https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/commit/291eabb1e844417990eda2661a382b493eb8e805)

> It's incredibly rare to see --yes (and similar) in installation instructions. Many users (myself included) prefer to preview the package manager transaction for new software before letting 'er fly. For automated scripts and build systems the non-interactive options are great, of course.

Yes, I noticed. However, I do not get the sentiment. If I find specific instructions for installing something, I expect them to be well curated or a bug otherwise. In both cases, I want to run them: In the first case to install, what I need and in the latter to get a log for a bug report. So why even interact with the package manager? Maybe I am privileged, because I could take the time curating [my own installation process since at least 2012](https://gist.github.com/Bengt/2937539/revisions?page=7). When I setup a new operating system like in the past days, I usually run through it twice: Once to run into all the new bugs, either with my guide or with the packages. I then file issues like this one and fix my guide. Finally, I reinstall the operating system one more time to a clean install with only minimal workarounds. After that I rarely install software anymore. I like to think that this makes workflow less uninterrupted by package installs, but I acknowledge that it might just be me being sensitive to the cumbersomeness that is package managers. - Especially, since I often forget the --yes flag and run apt in the background just to find it waiting for my [Enter] later.

[My current setup guide for reference](https://gist.github.com/Bengt/9)

Today, there is GitHub Actions which can test against virtually any operating system and thus instructions like these can actually be tested routinely. Changes to the README should ideally be executed under the relevant operating systems and changes to the relevant operating systems can be caught using a scheduled run of the respective job. So I do not see any excuse, why instructions should ever be wrong in the first place.

Also the author(s) of the package which requires additional installations using the package manager probably know better than me anyway which their software requires. I could only ever detect harsh faults like uninstalling large chunks of conflicting software or so. So largely, I as an end user of most of the software I install am dependent on the developers doing their part of providing software and installation instructions that are as close to hands-off as possible.

mgalgs:
> Yes, I noticed. However, I do not get the sentiment.

You make some good points. I personally still prefer to preview the transaction. Call it OCD, I don't know.

In any case, I'd still argue that consistency with the wider community is a more objective target.

darkxst:
I was probably the one that approved the --assumeyes patch, go ahead and remove that.
bit off-topic but why not use PackageKit over Dbus to trigger install of dependencies, gnome-control-center does that. It would probably work across Debian/Ubuntu and Fedora (with different package names) not sure about other distros

Bengt:
I added another commit which removes all the options forcing a non-interactive mode of the package managers. I guess consistency is a bit better at least, although this way around is not my preferred solution. Can this now be merged?

DevDef:
@chrisspen This PR should be closed now.

DevDef:
@chrisspen Let’s try to empty the PR queue... Please, close this one.

PR #647, opened 2020-11-21
Fix showing in top-left corner

Daemo00:
Fix https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/issues/633

Just a first attempt to fix the linked issue, mostly done by reverse-engineering the changes in https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1380 :sweat_smile:, but I tried locally and it fixes the issue for me.

PR #660, opened 2021-03-17
Bunch of fixes

DevDef:
This pull request contains various fixes such as:
- include the fix for the tooltip wrongly displayed at the upper left corner in GNOME 3.38 (#647)
- fix missing extension icon removed from Adwaita in recent versions of GNOME
- fix l10n of number format in disk usage
- add an update_template.sh script to extract the new strings to be translated
- run this script to update the PO translation files
- update the French translation


chrisspen:
Hi, thanks for your fixes. However, it looks like this is breaking the lint check. I assume the linter hasn't been updated to supported the most recent language version. Could you see if you can update the linter?

DevDef:
@chrisspen It seems that I’ve finally managed to fix the Travis build. :-)
Still based on the old Xenial though...

chrisspen:
Thanks. I'm also working on a version to upgrade it to Focal, hopefully using upstream Nodejs as well. If I can't get that to work, then I'll use what you've done, or at least adopt some of your Travis fixes.

PR #662, opened 2021-03-19
Typographic fixes + update and fix install instructions

DevDef:
Spell GNOME in uppercase instead of Gnome
Replace quotes by better ones
Add gnome-system-monitor to the list of packages needed
Add instructions to install on Mageia

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,9 +37,9 @@ Before installing this extension, ensure you have the necessary system packages 
 
       $ sudo urpmi lib64gtop-gir2.0 lib64nm-gir1.0 libclutter-gir1.0 gnome-system-monitor
       
-      or
+    or
       
-      $ sudo dnf install lib64gtop-gir2.0 lib64nm-gir1.0 libclutter-gir1.0 gnome-system-monitor
+      $ sudo dnf install lib64gtop-gir2.0 lib64nm-gir1.0 lib64clutter-gir1.0 gnome-system-monitor
       
 
 Additionally, if you have an NVIDIA graphics card, and want to monitor its memory usage, you'll need to install `nvidia-smi`.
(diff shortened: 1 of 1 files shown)


chrisspen:
I have to ask. Why did you replace all apostrophes with the non-standard `’` character? It's not a big deal, but I'm always frustrated when I see that since it's not simple ASCII.

DevDef:
Actually, you’re wrong, `’` is the true apostrophe and `'` is the neutral typewriter single quote:
https://en.wikipedia.org/wiki/Apostrophe
https://en.wikipedia.org/wiki/Quotation_mark#In_English
I think nowadays Unicode is safe and very well supported on every systems, so you shouldn’t feel frustrated. ;-)
ASCII is too restrictive, Unicode is the way to go…

p91paul:
He is not. Your own links claims the Unicode standard "prefers" U+2019 as apostrophe, but recognizes ' as the typewriter aphostrophe. ASCII has ', so it's not too restrictive here. Also, unicode might be mostly safe on Linux, so could be fine since this is a Linux extension, but different encoding on different systems is still a major headache.
So ' is still preferable, since it's present on all keyboards, ASCII, and noone uses the unicode one on text files anyway.

DevDef:
@p91paul 
Hi Paulo,

You’re out of topic, the README.md file goal is to generate the GitHub project page information. There’s no Web browser not supporting UTF-8 encoding. :-P

chrisspen:
>Actually, you’re wrong, ’ is the true apostrophe and ' is the neutral typewriter single quote

Yes, I'm aware it's part of the Unicode standard, but that's not the same thing as what's in standard use among developers, much less users at large.

It's best practice to keep README files simple and easy to read and maintain. They're not just viewed in web browsers, but in all manner of text viewers and editors. Requiring an editor with Unicode support, much less having to open a character map just to lookup and enter the special Unicode apostrophe and quote characters, when every user can simply press the ' key, makes little sense, especially when it's not adding anything to the document.

To that effect, please remove all the superficial Unicode changes from this PR. Let's keep the document plain ASCII for now.

DevDef:
Anyway, it’s not that important...
I think we/you should consider the other longtime pending PR, then wait for the translators to update their translations, and finally publish a new version on extensions.gnome.org in a few weeks.

PR #608, opened 2020-05-09
Updated Simplified Chinese translation

zhmars:

PR #667, opened 2021-03-22
Update Simplified Chinese translation

zhmars:
Ref: https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/pull/608#issuecomment-803634356


DevDef (approved):
Seems OK, but I don’t understand Chinese. ;-)

zhmars:
Thanks.

Issue #601, opened 2020-04-30
Logout when suspend or loock screen


buoncubi:
After the last update (I did it yesterday) I got a NaN issue with the disk IO. See [HERE](https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/issues/597#issuecomment-621886858) for my comments about it.
At the same time, I also realized that my ubuntu logs me out when try to *suspend* the PC or *lock the screen*. 
When I disabled the `system-monitor` extension it does not log me out anymore.

I have tried with [system-monitor](https://extensions.gnome.org/extension/120/system-monitor/) and [system-monitor-next](https://extensions.gnome.org/extension/3010/system-monitor-next/), but the behavior is the same. 

Any suggestions?
Thanks!

rebelinux:
i have the same logout issue but on archlinux

dseomn:
I think I'm having the same issue, though the shell restarts instead of logging me out. I think that's because I'm on Xorg. Are you on Wayland? Here are some logs:

```
May 08 10:58:26 solaria gsd-media-keys[15353]: Couldn't lock screen: GDBus.Error:org.freedesktop.DBus.Error.NoReply: Message recipient disconnected from message bus without replying
May 08 10:58:26 solaria systemd[15017]: gnome-shell-x11.service: Main process exited, code=killed, status=11/SEGV
May 08 10:58:26 solaria systemd[15017]: gnome-shell-x11.service: Failed with result 'signal'.
```

DZPM:
Same here: noticed some NaNs, and when I lock the computer, it crashes my session

This happened after a reboot (I don't reboot often), so it's possible the extension hadn't been reloaded in a while.

Work-around: I could lock again my session after disabling the extension.

I'm using Ubuntu 18.04 with kernel `5.3.0-51-generic`, and GNOME Shell `3.28.4`

enricorox:
I can confirm the problem but instead of log out gnome freezes and restart.
See my problem [here](https://askubuntu.com/questions/1235366/unable-to-lock-screen) with logs.

tangerrine-tj:
It seems that the gnome session crashed and restarted once the screen locked automatically or by hand.
When I disabled the system-monitor extension, no exceptions happened.

buoncubi:
[...]

Issue #597, opened 2020-04-27
Disk is Nan after last Update...


farhadinima75:
today, after install update, Disk unit in extension turns NaN. :(

how can i solve this?

Can i Downgrade update???

OS is ubuntu 18.04

eturkes:
Same here. He's a screenshot showing the situation along with the output of iotop and my system stats from screenfetch. Despite iotop showing no disk activity, the monitor is showing a maxed out value. And iotop is working correctly as it does show disk activity when that's actually going on, but no change in the system monitor applet.

![Screenshot from 2020-04-27 11-41-16](https://user-images.githubusercontent.com/25193231/80391908-6466c380-887c-11ea-8b1d-29ec65f30749.png)

nk7z:
SWAP is the same...  Not NAN, but zero...  Ubuntu.  Happened after last update.

AlexFalappa:
Already reported in #594

techsy730:
See https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/issues/594#issuecomment-620663485 for a fix for the disk NaN

chrisspen:
Everyone encountering this issue, please post your gnome-shell version, found with `gnome-shell --version`.

nk7z:
uname@nostromo:~$ gnome-shell --version
GNOME Shell 3.28.4

eturkes:
GNOME Shell 3.26.2

acesabe:
Same for me with Disk R/W, CPU Temp, Fan RPM - all NaN  

ASUS/AMD Phenom system
Ubuntu 18.04
GNOME Shell 3.28.4

Not the first time update has resulted in this issue...

buoncubi:
I have the NaN issue with disk IO only using
- GNOME Shell 3.28.4 
- Ubuntu 18.04
It solves by following the [comment posted above](https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/issues/597#issuecomment-620663833).

But this update leads to [another issue for me]().

manti-by:
Ubuntu 18.04
GNOME Shell 3.28.4

frankvw2017:
> Everyone encountering this issue, please post your gnome-shell version, found with `gnome-shell --version`.
GNOME Shell 3.28.4

nk7z:
GNOME Shell 3.28.4


73, and thanks,
Dave (NK7Z)
[...]

Issue #594, opened 2020-04-27
Not a number


vladimirnewissue:
Disk section shows "not a number" and filled with color. Screenshot below.

![notanumber](https://user-images.githubusercontent.com/64399031/80342895-a9372e00-886d-11ea-8413-43126162a83e.png)

farhadinima75:
same problem after today Update.... :(

AlexFalappa:
Can confirm this after update on 27 Apr 2020.

vladimirnewissue:
> Can confirm this after update on 27 Apr 2020.
> 
> I am on Ubuntu 18.04.
> 
> Any workaround?

me too ubuntu 18.04. no workaround yet...

ivocavalcante:
Same here, Ubuntu 18.04. Disk rates and Temp show NaN. Also, this is what temperature source selector shows:

![image](https://user-images.githubusercontent.com/11679725/80493479-d94d0280-893b-11ea-873d-ef6202581525.png)

techsy730:
For the Disk issue, I basically undid https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/commit/3ba2f60c45eac1fce8702387884f96840e224db2

Replace
`let lines = ByteArray.toString(as_r[1]).split('\n');`
with
`let lines = String(as_r[1]).split('\n');`

For the Temperature issue, the exact same fix described in
https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/issues/520#issuecomment-498990185 works here too.

vladimirnewissue:
> For the Disk issue, I basically undid [3ba2f60](https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/commit/3ba2f60c45eac1fce8702387884f96840e224db2)
> 
> In `~/.local/share/gnome-shell/extensions/system-monitor@paradoxxx.zero.gmail.com/extension.js`
> Find the section
> 
> ```js
>         let file = Gio.file_new_for_path('/proc/diskstats');
>         file.load_contents_async(null, (source, result) => {
>             let as_r = source.load_contents_finish(result);
>             let lines = ByteArray.toString(as_r[1]).split('\n');
> ```
> 
> Replace the last line:
> 
> ```js
> let lines = ByteArray.toString(as_r[1]).split('\n');
> ```
> 
> with
> 
> ```js
> let lines = String(as_r[1]).split('\n');
> ```
> 
[...]

Issue #592, opened 2020-04-27
GPU usage reports NaN for glxinfo output


psukys:
The used/total  is shown correctly, whereas the usage percentage is reported as NaN (see screenshot for example).

![image](https://user-images.githubusercontent.com/1302529/80339474-62463a00-8867-11ea-84b0-fec2b580a98f.png)

farhadinima75:
hi @psukys 
for showing fan speed in panel, what did you do ?

I've already installed lm-sensors but fan speed and thermal doesn't work and show NaN....

**Second problem**:  after today update, my Disk is showing NaN like your problem.

my OS : Ubuntu 18.04

happysmack:
I had the snap version for ubuntu installed and most worked.  I do not like snaps so I removed and put the repo version back like I always had before. Now the GPU ( Nvidia ) just shows a value for memory. I have the NAN% for the other GPU setting, disk activity, thermal, and fan.  I have lm-sensors and the files listed in the readme installed.  I know at least the disk activity did work when I first installed 18.04.

RolandAckerl:
I had the same issue as psukys, Ubuntu 20.04, CUDA 10.2, Nvidia-smi 440.64, untamed potato-power of gtx 960M ;)


Just to make it clear: afaik only Nvidia GPUs are supported.
Check if their drivers are loaded properly (easy way in terminal: `nvidia-smi`).
If, like in my case, you are running your system on Intel's built-in GPU but other stuff with Nvidia (for e.g. CUDA or similar), run `nvidia-settings`, pick **PRIME Profiles** menu in the window and enable **NVIDIA On-Demand**. Like:
![image](https://user-images.githubusercontent.com/32834323/81702484-a2aed600-946b-11ea-96b2-e0aa1d0ae11a.png)

Then reboot.

worked for me (GPU not in use here):
![image](https://user-images.githubusercontent.com/32834323/81701726-b3ab1780-946a-11ea-8799-f45fcd999f42.png)
running a small neural net (MNIST-training):
![image](https://user-images.githubusercontent.com/32834323/81702364-75fabe80-946b-11ea-9351-3520c7067e6e.png)

PR #674, opened 2021-04-06
add backward support for Gnome 3.28 , fix NaN temperature and disk IO…

bluet:
… on Ubuntu 18.04

add backward support for Gnome 3.28 , fix NaN temperature and disk IO on Ubuntu 18.04 . 

related issues: #520 #526 #594 #597 #601  https://github.com/thankjura/ds4battery/issues/4  
I don't have dedicated GPU to test #592 but it should be possible to fix in the same way.

inspired by @shemgp and @EBoisseauSierra

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -107,6 +107,13 @@ Clutter.Actor.prototype.reparent = function reparent(newParent) {
     newParent.add_child(this);
 }
 
+function parse_bytearray(bytearray) {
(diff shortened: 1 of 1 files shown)


bluet:
Result (3.28)
2021-04-06 17-45-04 的螢幕擷圖

chrisspen:
Thank you.

Tutorgaming:
in my machine, it still shows NaN as output (Thinkpad T490 + Ubuntu 18.04 + Gnome Shell 3.28.4)
But when I use the combination between Math.round and parseFloat (instead of parseInt) it works as it should be.

I'm not sure if it is machine-dependent or not.

Issue #633, opened 2020-10-11
Tooltips showing in top-left screen corner


appleynski:
![image](https://user-images.githubusercontent.com/3051283/95677633-0d08a580-0bc7-11eb-88ef-889b8917e4e0.png)

When hovering mouse pointer over any graph, the tooltip always shows up in the top-left corner of the screen.
I am running GNOME 3.38.1 on Xorg on Arch Linux with nvidia proprietary drivers.

Awesome extension by the way!

dpriedel:
I see the same thing running Gnome 3.38.1 on Manjaro with the nouveau driver.

franglais125:
This is due to the fact that the method we use to place the dialog correctly was removed upstream, and replaced with a newer one [1].

Namely `util_get_transformed_allocation` is no longer available.

We need to update the code for this.

[1] https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1380

darkxst:
that tooltip code is ancient and ugly, surely there are nicer ways to do it now ?

franglais125:
@darkxst thanks for chiming in. I will try to work on this (and many other issues), as soon as possible. Unfortuntely, that probably means some time around January. The good news is that this is not a deal breaker (i.e. the extension doesn't crash on 3.38).

Another thing I have in mind that needs improving is HiDPI support. That shouldn't be so bad to implement, just a few things to resize using the scale factor. I have this running locally.

darkxst:
@franglais125 my laptop is HiDPI 4k screen, I dont use this extension these days, but happy to test your patches if that helps.

FastThenLeft:
Same problem with Ubuntu 20.10 (Gnome 3.38.1).

shemgp:
Fixed in https://github.com/Daemo00/gnome-shell-system-monitor-applet/commit/e1d1e8a4cfe6f86db47d22752e748a0bb7d79859

bluet:
I'm experiencing this bug with Ubuntu 18.04 (Gnome 3.28).

@shemgp seems that @Daemo00 's https://github.com/Daemo00/gnome-shell-system-monitor-applet/commit/e1d1e8a4cfe6f86db47d22752e748a0bb7d79859 is not merged and not belong to any PR?

bk138:
I am still having this issue, mutter 3.38.4, system-monitor 39, _only when running under Wayland_.

PR #675, opened 2021-04-06
Make tooltips show under the graph instead of the left of the screen

shemgp:
Fixes: #633 by using `get_transformed_extents` only on mutter where it exists.

Tested on Ubuntu 18.04 and Ubuntu 21.04.


chrisspen:
Thanks.

PR #676, opened 2021-04-06
Fixed thermal sensor dropdown label rendering.

chrisspen:
Fixed rendering of the thermal dropdown text in Gnome ~3.28.4 inspired by PR #674.

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/prefs.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/prefs.js
@@ -38,6 +38,13 @@ function color_to_hex(color) {
     return output;
 }
 
+function parse_bytearray(bytearray) {
+    if (!ByteArray.toString(bytearray).match(/GjsModule byteArray/)) {
+        return ByteArray.toString(bytearray);
+    }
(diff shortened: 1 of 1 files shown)


bluet:
@chrisspen  Cool. I was experiencing the problem with all drop-down labels but couldn't find where to fix it.

Thanks for your fix! 🎉

bluet:
@chrisspen could you also fix other drop-down menus this way in the PR?

chrisspen:
@bluet The thermal sensor dropdown is the only one broken for me in Ubuntu 18. What other ones are broken?

I didn't already merge this because I'm not sure it doesn't break anything in more recent Gnome versions.

bluet:
@chrisspen the Fan one seems broken too
2021-04-07 21-23-36 的螢幕擷圖

chrisspen:
@bluet It works for me. Which OS/Gnome version are you using?

bluet:
@chrisspen just did another test again with the latest commit, it works list a charm!

I can confirm that it works with my Gnome 3.28.4 on Ubuntu 18.04

Issue #672, opened 2021-04-01
Add support for GNOME 40 and an updated version to https://extensions.gnome.org/extension/120/system-monitor/


Kerrung:
Add support for GNOME 40 and an updated version to https://extensions.gnome.org/extension/120/system-monitor/

p91paul:
The extension mostly work by just changing version in metadata.json.
Preferences don't, of course, because of the gtk 4 switch.

BlackDex:
Using Arch Linux here.

Kerrung:
> What distro/release is everyone mostly using?

I use Arch Linux (rolling release).

ludenticus:
> What distro/release is everyone mostly using?

Arch Linux user too

leogr:
Arch Linux user

d2r2:
> The extension mostly work by just changing version in metadata.json.
> Preferences don't, of course, because of the gtk 4 switch.

Thanks @p91paul! Helps as a temporary fix, if append "40" to "shell-version": [...] in metadata.json.

Arjdroid:
> What distro/release is everyone mostly using?

Arch Linux (Rolling Release)

I would assume most users with this issue will be Arch Linux or other Rolling Release distro users because other distros like Pop!_OS and Ubuntu are still on GNOME 3.8 with the major exception being Fedora

Arjdroid:
> > The extension mostly work by just changing version in metadata.json.
> > Preferences don't, of course, because of the gtk 4 switch.
> 
> Thanks @p91paul!
> To All. As a temporary partial fix, append `"40"` to `"shell-version": [...]` in metadata.json.
> P.S. The extension is usually located in the folder: `~/.local/share/gnome-shell/extensions/system-monitor@paradoxxx.zero.gmail.com/`

Would I be correct in assuming that this workaround only works for users who already had this extension on their system BEFORE "upgrading" to GNOME 40 because I am on a GNOME 40 Arch Linux system and have tried this workaround to get the manually installed version working but the Extensions App doesn't allow me to switch on the extension and as you have mentioned the Preferences (gear icon button in my case) is not working and give a GTK-related error.

Is there anything I could try to trick the system into running the extension like some sort of file changes and then making Preference changes via some file changes like copying the configuration file from an existing Pop!_OS system running the extension with my desired configuration into the directory where it checks for my preferences? 

[...]

PR #678, opened 2021-04-08
update prefs to support gnome 40

kgarner7:
Hopefully addreses #672 

Gtk4 introduced breaking changes to how the preferences are created. This PR seeks to handle those changes while maintaining backwards compatibility
-  Gtk.HBox and Gtk.VBox are dead, replaced with Gtk.Box with orientation 
- `border_width` is dead, replace it with margins of 10 (this seems to function similarly)
- `add`, `pack_start`, `pack_end` calls for Box have been replaced with `append` and `prepend`: these calls are aliased where appropriate
- Because packing is gone, spacing is now controlled by `set_homogeneous`; It has bigger windows and boxes, but seems to be the closest to maintaining packing?
- `show_all` is unneeded in new versions
- `get_all_children` is gone, so I use effectively an iterator to build and sort the array

Additionally fixes complaints in `extension.js` from eslint.

Tested on fedora 34, arch linux (in English, for ordering).

I'm not sure if this is the best/cleanest way to multiplex, but hopefully it is at least a good start?

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/prefs.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/prefs.js
@@ -69,39 +69,52 @@ function check_sensors(sensor_type) {
     return [sensor_list, string_list];
 }
 
-function box(isHorizontal, spacing, hasBorder) {
+/**
+ * @param args.hasBorder Whether the box has a border (true) or not
+ * @param args.horizontal Whether the box is horizontal (true)
+ *      or vertical (false)
+ * @param args.shouldPack Determines whether a horizontal box should have
+ *      uniform spacing for its children. Only applies to horizontal boxes
+ * @param args.spacing The amount of spacing for a given box
+ * @returns a new Box with settings specified by args
+ */
+function box(args = {}) {
     const options = { };
 
-    if (typeof spacing !== 'undefined') {
-        options.spacing = spacing;
+    if (typeof args.spacing !== 'undefined') {
+        options.spacing = args.spacing;
     }
 
     if (shellMajorVersion < 40) {
-        if (hasBorder) {
+        if (args.hasBorder) {
             options.border_width = 10;
         }
 
-        return isHorizontal ? new Gtk.HBox(options) : new Gtk.VBox(options);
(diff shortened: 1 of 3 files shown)


system-monitor@paradoxxx.zero.gmail.com/prefs.js
```diff
@@ -66,12 +69,49 @@ function check_sensors(sensor_type) {
...
+
+    return aliasBox;
+}
 
 const ColorSelect = class SystemMonitor_ColorSelect {
     constructor(name) {
         this.label = new Gtk.Label({label: name + _(':')});
         this.picker = new Gtk.ColorButton();
-        this.actor = new Gtk.HBox({spacing: 5});
+        this.actor = box(5);
```
p91paul, line 114: this one should have been box(true, 5) probably!

kgarner7: Thanks for that catch!

p91paul:
Tested this branch, it mostly works, however the preferences window doesn't fit anymore in my 1366x768 laptop screen.
Perhaps you exaggerated some spacing?

p91paul:
My advice, merge this, release to e.g.o., then fix the small annoyances once you stopped the flood of complaining users.

kgarner7:
The spacing issue was caused by the color selects being expanded to a larger size when `homogeneous` was set to true; this consequently scaled up the size of all other horizontal boxes. 

I've modified `box` to conditionally apply spacing, so it should no longer have that issue. Unfortunately this makes colors left-aligned instead of center aligned, but it should work on 1366x768 and 1280x (preserving packing for rows). It is too large for 1024x, which would require different spacing/packing options.

p91paul:
Tested, looks good enough for me.
The window has always been pretty large, so probably 1024x users (if any) won't actually see a regression.

kgarner7:
Sorry for delayed response. I think it should be stable

Digitalone1:
If it's good, please merge.

chrisspen:
I'm testing locally.

PR #681, opened 2021-04-19
Add GitHub CI pipeline

d97jro:
GitHub's CI pipeline makes it very simple to rapidly set up powerful checks for one's code tree.
This PR adds such a pipeline to this project, beginning with the GitHub shellcheck action.
(Currently set to only fail on errors, and not the default behavior to fail on warnings too since there are a handful of scripts that currently do not pass shellcheck as-is with shellcheck's default behavior.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,19 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  # For more information on GitHub's ShellCheck Action:
+  # https://github.com/marketplace/actions/shellcheck
+  # For more information on shellcheck failures:
+  # https://github.com/koalaman/shellcheck/wiki/Checks
+  shellcheck:
+    name: shellcheck
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


chrisspen:
Interesting. I haven't experimented with Github's CI yet. Have you compared this against our current Travis-CI config in `.travis.yml`?

d97jro:
> Interesting. I haven't experimented with Github's CI yet. Have you compared this against our current Travis-CI config in `.travis.yml`?

That totally went past me and I did not see that. Answer: no, I have not. :-)
But I guess this is the point when I get to know more about Travis.

With a GitHub-hosted project, I would find it appealing to leverage the available GitHub functionality (just as the GitLab ditto when running on GitLab), but I guess there are good reasons behind using Travis.

If you would like me to, I could have a go at putting the shellcheck into Travis instead?

chrisspen:
As I've gotten the Docker build to run in Travis-CI, and it looks like getting Docker to work correctly in Github may be non-trivial, I think I'm going to shelve the Github CI for now, unless someone can get the `docker/run_tests.sh` working in Github CI.

PR #684, opened 2021-04-21
Add GUI testing proof-of-concept

### LINKED ISSUE #663: Create UI integration test proof-of-concept
Use a tool like [Dogtail](https://gitlab.com/dogtail/dogtail) or similar to script integration tests that can be run inside Docker.

The first test should be able to do something very simple like:

1. install the extension
2. detect the location of the graphical widget in the top panel and click it
3. confirm an overlay appears
4. confirm a label containing the text "Cpu" exists, with a value containing text in an integer format.

Chocobo1:
Closes #663.

The [dogtail](https://gitlab.com/dogtail/dogtail) test tool is used to write the tests as suggested in the linked issue.
The tests run inside the docker image, steps:
```shell
# build new image
./build-docker.sh

# startup
./run-docker.sh
./open-docker.sh

# running GUI tests
# open terminal within docker container and then
cd /home/default/shared
./run_gui_tests.sh

# shutdown
./close-docker.sh
```

### DIFF
--- a/Dockerfile.ubuntu2004
+++ b/Dockerfile.ubuntu2004
@@ -102,5 +102,14 @@ RUN mkdir -p $HOME/.vnc
 COPY xstartup $HOME/.vnc/xstartup
 RUN echo "password" | vncpasswd -f >> $HOME/.vnc/passwd && chmod 600 $HOME/.vnc/passwd
 
+# Set up GUI testing environment
+RUN sudo apt-get update \
+    && sudo apt-get install -y python3-pip \
+    && sudo apt-get clean -y \
+    && sudo rm -rf /var/lib/apt/lists/*
+RUN pip3 install --user dogtail
+RUN chmod +x "/home/${USER}/.local/bin"/*
+RUN echo "export PATH=\"\$PATH:/home/${USER}/.local/bin\"" >> "/home/${USER}/.bashrc"
+
 # Switch back to root to start systemd.
 USER root
--- a/run-docker.sh
+++ b/run-docker.sh
@@ -3,5 +3,6 @@ docker run -it --name=ubuntu-gnome --rm \
     --tmpfs /run --tmpfs /run/lock --tmpfs /tmp \
     --cap-add SYS_BOOT --cap-add SYS_ADMIN \
     -v /sys/fs/cgroup:/sys/fs/cgroup \
+    -v "$(pwd)/testing:/home/default/shared" \
     -p 5901:5901 -p 6901:6901 \
     gnome-shell-system-monitor-applet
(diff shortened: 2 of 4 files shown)


chrisspen:
Thanks. This looks very interesting. I'm glad you responded to my post on bountysource. I'm testing this now.

The only minor issue I see is this still requires a user to manually login to the desktop to run the test script. I ran into that limitation myself when building out the Docker framework.

However ideally, I'd like to run this from inside a headless CI system with no user interaction.

Do you see any easy way to get this working without having to call open-docker.sh? Maybe changing CMD in the Dockerfile to use screen to launch both `/sbin/init` and a custom runner for your `run_gui_tests.sh`? Although there still might be problems with logging in the user.

chrisspen:
This looks great. Works exactly as described. Great job.

It still requires a user to login, which isn't ideal, but that wasn't part of my original requirements, so I won't hold you to that.

@paradoxxxzero @franglais125 @DevDef Any thoughts before I merge this? I'd like to flesh out some formal automated integration testing in our CI, and programmatically accessing the Gnome UI was the last big missing piece.

Chocobo1:
>Do you see any easy way to get this working without having to call open-docker.sh? Maybe changing CMD in the Dockerfile to use screen to launch both /sbin/init and a custom runner for your run_gui_tests.sh? 

Beats me, can't say I have many experiences with docker and headless systems.
AFAIK, generally you'll need to go through the following steps:
1. Allocate a virtual screen
  Something like Xvfb might be of help here if it is running X11 server. Also note that the testing scripts are written/tested under the default 1368x768 resolution.
2. Login the default user
  (Maybe use auto-login?)
3. Run the testing scripts
  (No idea how to trigger this...)

>This looks great. Works exactly as described. Great job.

You're welcome! :)

I felt that I'll need to talk a little about writing the test script. It is near impossible to achieve this without the help from a tool called `sniff` (which is install along with dogtail). When writing the script, I fire up `sniff` and start to find the target fields, you can see how it matches the `Cpu` field in the following screenshot:
screenshot
As you can see, most of the field names are just empty and `sniff` doesn't provide a way to search the whole tree. Thus it requires a bit work to find the targets manually. After that, the rest is just writing code for extracting the values and pattern matching the field's `Name:` and `Role name:`.

In some way, I felt the testing script is quite brittle as it depends on the exact layout of the widgets and maintaining such scripts would require non-trivial efforts if the layout changes often (I hope that is not the case for this project).

chrisspen:
@Chocobo1 Thanks, that's very helpful. Yeah, I played around with dogtail a little, but identifying the elements in Gnome's DOM to click was where I ran into a roadblock. I agree that the script looks a little fragile. Fortunately, our UI is pretty stable, even if we're constantly tweaking things in the backend.

PR #552, opened 2019-11-27
Wrap into register class. Conditionally decode ByteArrays from files.

ettavolt:

PR #686, opened 2021-04-24
Wrap into register class. Conditionally decode ByteArrays from files.

ettavolt:
What's left of #552.


chrisspen:
This change appears to have completely broken the applet on Ubuntu 18.

Issue #613, opened 2020-05-21
On HiDPI displays, pie disk utilization is not centered


bludhemn:
![gnome-system-monitor](https://user-images.githubusercontent.com/19481454/82535832-99281b00-9b9b-11ea-87fa-bba8b4fd33de.png)

Issue #525, opened 2019-05-18
hi-dpi small scaling


fakhamatia:
In 4K monitor on 3200 resolution its show small
Graph is bigger on HD resolution , but text its very small

![Screenshot from 2019-05-18 11-36-30](https://user-images.githubusercontent.com/20030910/57965984-32a56700-7961-11e9-93b7-842e9c429a93.png)

PR #687, opened 2021-04-24
Add support for display scale factor

pitdicker:
Not all parts of the interface follow the display scale setting on a high dpi monitor. Text is scaled automatically, but the graphs, pie chart and bar chart remained tiny.

This is my attempt to have those elements use the right scale factor. It is only read once, you have to restart the extension or Gnome Shell after changing the monitor scale factor.

(Also, this is my first time touching Javascript and Gnome Shell extensions.)

This should fix https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/issues/525, although that issue is a bit vague. Fixes https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/issues/613.

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -398,21 +398,17 @@ const Chart = class SystemMonitor_Chart {
         }
         cr.$dispose();
     }
-    resize(schema, key) {
-        let old_width = this.width;
-        this.width = Schema.get_int(key);
-        if (old_width === this.width) {
+    resize(width) {
+        if (this.width === width) {
             return;
         }
-        if (Style.get('') === '-compact') {
-            this.width = Math.round(this.width / 1.5);
-        }
-        this.actor.set_width(this.width);
+        this.width = width;
         if (this.width < this.data[0].length) {
             for (let i = 0; i < this.parentC.colors.length; i++) {
                 this.data[i] = this.data[i].slice(-this.width);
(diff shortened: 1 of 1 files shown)


pitdicker:
I'll try to add a callback for when the scaling factor changes so it just works, like the rest of the shell. Maybe do so for changing the compact setting and the text scale factor as well, as those need changes in roughly the same place.

Please wait with review  for now.

pitdicker:
This should now be ready.

There are a number of changes, those related to HiDPI:
- `Chart` and `Graph` are now both aware of the display scale factor, and text scaling.
- Charts now exactly line every sample to a pixel, which means using half pixel values. See https://www.cairographics.org/FAQ/#sharp_lines
- Charts use bilinear interpolation between samples. Previously spikes were 0px wide at the top, and 2px at the bottom. Now they are 0.5px wide at the top and 1.5px at the bottom. This makes the true top better visible, and gives less triangular spikes.

Issues I happened to run in to while figuring out how to add a scale factor:
- Pie and Bar graphs didn't scale with text scaling. Because the menu is wider with larger text anyway, I think it is better to scale them.
- If there are many mounts, the text in Pie can start to overlap. Now also scale the text if the lineheight factor is less than 1.2.
- if a mount is more than 75% full, it can start to overlap with the text. Draw the text after drawing the rings, so the text stays on top.

Sorry for the large PR. I tried to split things into reasonable commits. Let me know what you think!

pitdicker:
Thank you!

PR #693, opened 2021-05-17
prefs: Remove clutter dependency on GNOME 40

jtojnar:
There is a strict distinction between GNOME Shell using the Clutter toolkit and apps (including extension preferences) using GTK.

And while it is technically possible for apps to use Clutter, the library is unmaintained and its use is discouraged. Additionally, GNOME Shell actually uses its own internal fork of Clutter so if we depend on Clutter, we need to install a second copy of it just for the preferences.

Fortunately, GDK 4 [supports parsing hex colours with alpha values](https://gitlab.gnome.org/GNOME/gtk/merge_requests/2356), so we can just use it directly without depending on Clutter.

We will only use the Compat module when GTK 4 is not available, such as on GNOME < 40.

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/prefs.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/prefs.js
@@ -9,7 +9,6 @@ const Gettext = imports.gettext.domain('system-monitor');
 
 let extension = imports.misc.extensionUtils.getCurrentExtension();
 let convenience = extension.imports.convenience;
-let Compat = extension.imports.compat;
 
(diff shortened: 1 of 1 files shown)


bjornfor:
I tested this PR in a NixOS 21.05 VM. Success! (I can now open "Preferences" without a crash.)

bjornfor:
@chrisspen: Is there anything I can do to get this landed?

bjornfor:
@jtojnar and @chrisspen : Thanks!

PR #696, opened 2021-06-14
Require clutter for Arch Linux

auipga:
Without clutter I couldn't open the preferences
Fixes: `Error: Requiring Clutter ...`
Tested with master branch and GS 40


bjornfor:
Merging https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/pull/693 will drop the dependency on clutter for GNOME >= 40, so this one can be reverted.

PR #698, opened 2021-06-16
Fix Travis CI docker hang

braewoods:


braewoods:
It now "succeeds" as far as Travis CI is concerned. See here: https://travis-ci.com/github/braewoods/gnome-shell-system-monitor-applet/builds/229275759

chrisspen:
Thanks. If the chown command is the only fix, could you remove all the other unrelated JS changes? They're probably fine, but I'd like to keep this PR isolated to the CI build, and not address any other issues.

braewoods:
@chrisspen Ok. It's done. I've isolated this to just my commit.

chrisspen:
Thanks.

PR #700, opened 2021-07-06
Added the possibility to rearrange the graphs

### LINKED ISSUE #245: Order of items changed

DColadas:
Updates #612 to fix the conflicts. From that same PR:

> 
> I added some XML attributes and changed some code in enable() and the Settings window, making it possible to reorder the graphs (shown in the status bar and the information window) and the settings in the Settings window. This fixes #245 and is a nice feature.
> 
> To start rearranging, go to Settings, grab the tab label you wish to move, and drag it to its position. The changes in the order of the tabs are immediate (they will remain the next time the window is opened), but the displays in the status bar will only move after restarting GNOME Shell (Alt+F2 and the command r), although I may be able to work on improving it. The default order is the same as the previous one.
> 
> If you have some improvements (a better option than drag and dropping, for instance) or suggestions, feel free to comment!
>


system-monitor@paradoxxx.zero.gmail.com/schemas/org.gnome.shell.extensions.system-monitor.gschema.xml
```diff
@@ -458,6 +458,46 @@
 	  <default>false</default>
 	  <summary>Hide system battery icon</summary>
 	</key>
+  <key name="cpu-position" type="i">
```
ZimbiX, line 461: I reckon change the indentation of these new settings to match the others. Might as well fix up the indentation of the other elements too while we're at it.

ZimbiX (approved):
This is very cool! Well done =)

I think the tab dragging is a great way to do this, and I think is probably discoverable enough.

Without the order update taking immediate effect though, it could be good to add a note somewhere that a reload is required after changing it.

Tested rebased on https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/pull/754.

Between tests, I used this to reset the position settings:

```bash
dconf dump /org/gnome/shell/extensions/system-monitor/ | grep position | cut -d= -f1 | xargs -I '{}' -- dconf reset '/org/gnome/shell/extensions/system-monitor/{}'
```

PR #708, opened 2021-08-05
Updated Brazilian Portuguese and Portuguese

lucassfank:


po/pt_BR/system-monitor.po
```diff
@@ -303,7 +303,7 @@ msgstr "Estilo de uso"
 
 #: prefs.js:436
 msgid "pie"
-msgstr "setores"
+msgstr "pizza"
```
chrisspen, line 307: Are you sure about this? "Pie" in this context refers to a circular graph. Not the food.

chrisspen (changes requested):

po/pt_BR/system-monitor.po
lucassfank: Hi! Yes, here in Brazil this type of graph is called "Gráfico de pizza"

PR #709, opened 2021-08-11
Word For Word translation & better translation

TerraXD:
I fixed every word for word translation (otherwise it doesn't really mean anything)
and did some minor tweaks to help Novices/Begginers Understand a bit better.

### DIFF
--- a/po/fr/system-monitor.po
+++ b/po/fr/system-monitor.po
@@ -33,7 +33,7 @@ msgstr "Rafraîchissement"
 
 #: prefs.js:362
 msgid "Graph Width"
-msgstr "Largeur du graphe"
+msgstr "Largeur du graphique"
 
 #: prefs.js:348
 msgid "Show Text"
@@ -41,7 +41,7 @@ msgstr "Afficher le texte"
 
 #: prefs.js:367
 msgid "Display Style"
-msgstr "Type d’affichage"
+msgstr "Style d’affichage"
 
 #: prefs.js:490
 msgid "Display Icon"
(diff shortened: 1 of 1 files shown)


kubrickfr:
I'm a French native speaker and I disagree with this pull request

- Largeur du graphe: "graphique" is a bit more appropriate, I agree, but if we have to be pedantic it's a "diagramme" ("diagramme en bâtons empilés"). 
- Type d’affichage. That's debatable, but not worth the change, both are correct. Style is a better word for word translation but type is better in context
- chiffres: the English should be changed to "digits" or "digital", but the French should be "Chiffres" or "Digital", certainly not "Chiffre"
- tr/min, this is the [correct abbreviation](https://fr.wikipedia.org/wiki/Tour_par_minute). Nothing to change here,

chrisspen:
@kubrickfr Feel free to submit a PR with corrections.

kubrickfr:
Well, I'm not an authority :-D let's see if @TerraXD has something to say...

TerraXD:
I agree with that, i should be more sérious doin this

chrisspen:
@kubrickfr If you're a native French speaker, than you more of an authority than most. I'm certainly not a native French speaker, nor do I have any real ability to review translations. We rely on user submissions and reviews like yours to make sure they make sense. And ultimately, you're the one who's going to be using the translation.

kubrickfr:
all right, then I'll go down the rabbit hole and review the whole translation, and propose a changeset that _I_ deem consistent. But I will really appreciate reviews, language is usually more nuanced than right or wrong...

TerraXD:
> all right, then I'll go down the rabbit hole and review the whole translation, and propose a changeset that _I_ deem consistent. But I will really appreciate reviews, language is usually more nuanced than right or wrong...

Bonne chance mon pote

PR #715, opened 2021-09-23
Update README.md

dcerisano:
Resolve "ERROR"  on Ubuntu.


chrisspen:
You shouldn't have to reboot, and I never have when installing this extension on Ubuntu. What error did you encounter?

PR #716, opened 2021-09-27
Change ordering to be consistent with standard perf order.

purpleidea:
This restores the original order that was present in the classic GNOME
2.x system monitor that inspired this extension. Of course if this order
  was easily configurable by the user, that would be preferable.


DColadas:
#700 should already allow the user to choose the order

chrisspen:
If this change is still relevant, could you please resolve conflicts?

PR #718, opened 2021-10-01
Support Gnome 41.

### LINKED ISSUE #717: Support GNOME 41
After updating to Gnome 41, I see the following error:
`Gtk.VBox is not a constructor`. Which is likely related to fack that `VBox` is deprecated:
https://docs.gtk.org/gtk3/class.VBox.html

marxin:
Tested on openSUSE Tumbleweed.
Fixes #717.

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/metadata.json
+++ b/system-monitor@paradoxxx.zero.gmail.com/metadata.json
@@ -1,5 +1,5 @@
 {
-    "shell-version": ["3.26", "3.28", "3.30", "3.32", "3.34", "3.36", "40"],
+    "shell-version": ["3.26", "3.28", "3.30", "3.32", "3.34", "3.36", "40", "41"],
     "uuid": "system-monitor@paradoxxx.zero.gmail.com",
     "name": "system-monitor",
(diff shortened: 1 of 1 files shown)


jhaiduce:
This seems to be fixed in the master branch, but it looks like the fix hasn't reached extensions.gnome.org yet. Plus Gnome Extensions seems to be getting more aggressive about updating extensions, so when I install from master it keeps getting replaced by the version at extensions.gnome.org.

Can someone push the fix to extensions.gnome.org so we can just install from there?

In the meantime does anyone know a way to keep Gnome Extensions from "upgrading" the extension to the version at extensions.gnome.org?

p91paul:
you can manually set "version" to 1000 or some other high number in metadata.json, to make e.g.o. think you are on a higher version

jhaiduce:
> you can manually set "version" to 1000 or some other high number in metadata.json, to make e.g.o. think you are on a higher version

That helped, thanks.

190n:
Is there an ETA on when the GNOME 41 version will be published to extensions.gnome.org?

marxin:
> Is there an ETA on when the GNOME 41 version will be published to extensions.gnome.org?

Very good question. I guess @chrisspen can help us?

PR #723, opened 2021-12-05
Fix for disk speed problem

JuniorPolegato:
Issue: paradoxxxzero#489
The original code was accumulating sectors of disks/block devices and partitions, duplicating the read and write sectors counting.
Now the code gets the block device list and just accumulates their sectors.
I found a division of these sectors (512 bytes each) by 1024 and 8 (8192 => 8k) to get MiB, but that's not right, it is needed to divide by 2k (or *512/1024/1024) not 8k, maybe due to the accumulation of block devices and partitions, however, when operating on one partition (dd if=/dev/sda1 of=/dev/null bs=4k), it duplicates the accumulated sectors, so you need to divide by 4k, not 8k, getting half the speed. But if you are operating directly on a disk device (dd if=/dev/sda of=/dev/null bs=4k), does not duplicate the accumulated sectors, so you need to divide by 2k, not 8k, getting half of half the speed.
Nowadays, disk space is in MB, not in MiB, so I adjusted this too.

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -1495,7 +1495,7 @@ const Disk = class SystemMonitor_Disk extends ElementBase {
         this.last = [0, 0];
         this.usage = [0, 0];
         this.last_time = 0;
-        this.tip_format(_('MiB/s'));
+        this.tip_format(_('MB/s'));  // changed from MiB to MB (space standard)
         this.update();
     }
     update_mounts(mounts) {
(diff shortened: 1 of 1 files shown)


NVieville:
Hello,

> Nowadays, disk space is in MB, not in MiB, so I adjusted this too.

Maybe this could be discussed without raising a troll (very old IT debate).

According to [1] usage of the units tends to be as this: "base 2 values for memory and disk space values, and base 10 values for space on physical hard drives".
But this is also not so clear according to [2].
There are also a lot of other WEB pages debating this subject.

Maybe, here, the user point of view should be the purpose of printing values and units.
 
In case 1, if the values measure the amount of files transmitted, it should be kept in MiB. This corresponds to values printed by OSes when dealing with files size, and usually matches what users are most proficient in. Differences between MiB and MB are in general difficult to apprehend for final users.

In case 2, if the values measure the hardware throughput of the hard disk, it should be printed in MB. This corresponds to values printed when measuring a flow of bytes. OSes administrators, hardware maintainers or designers sometimes need these units, even if iops tends to be preferred by some.

This is really a matter of usage.
One thing that could be great is to let the user choose in what units to print values between MB and MiB, but this is another story and a different implementation.

These comments should only be taken for what they are: my 2 cents and a way to think about it.
Hope this will help. Any comments are welcome. 

Cordially,


-- 
NVieville

[1] https://www.ibm.com/docs/en/storage-insights?topic=overview-units-measurement-storage-data
[2] https://physics.nist.gov/cuu/Units/binary.html

chrisspen:
This is failing the build due to some Javascript formatting errors caught by checkjs.sh.

PR #736, opened 2022-02-04
I can confirm that it works for GNOME 42~alpha

shemgp:
Tested using my GNOME ppa with version gnome-shell 42~alpha and it works.


okkymabruri:
@paradoxxxzero please accept

PR #748, opened 2022-08-05
Gnome 3.2

kaliAxmed:
pls update to gnome 42


chrisspen:
I don't really see how these changes are necessary. The metadata.json already lists more recent versions than Gnome 3.2. And why remove the link to extensions.gnome.org? That's still a valid install path.

Closing as unnecessary. Feel free to add comments if you see additional reasons for these changes.

PR #752, opened 2022-09-19
Fix a small typo in extension.js

GaryElshaw:
Fix a small typo in the Disk menu. There was a missing space between /s and R.

PR #753, opened 2022-09-22
Add space before 'R' in disk usage string

Layerex:


GaryElshaw:
Great minds....Fools never differ? https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/pull/752

Issue #749, opened 2022-08-26
Gnome 43


galen1423:
Hi,

Is possible to update this extension to G.S 43? Currently, it produce this error:
`TypeError: StatusArea.aggregateMenu is undefined`

Thanks for your work :+1:

porjo:
I was able to make it work by adding `43` to `shell-version` array in `metadata.json` and also commenting out this line from `extension.js`:
```
  Main.__sm.elts.push(new Battery());
```
It means you won't be able to display Battery status :cry:

galen1423:
Hi @porjo,

Thank you for your workaround. I don't have any battery on my computer so, no problem. :-)

I hope this project will continue to be maintained.

marxin:
Yes, the following workaround works:

```diff
diff --git a/system-monitor@paradoxxx.zero.gmail.com/extension.js b/system-monitor@paradoxxx.zero.gmail.com/extension.js
index 1c56f0d..d75b77f 100644
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -2448,7 +2448,7 @@ function enable() {
         Main.__sm.elts.push(new Gpu());
         Main.__sm.elts.push(new Thermal());
         Main.__sm.elts.push(new Fan());
-        Main.__sm.elts.push(new Battery());
+//        Main.__sm.elts.push(new Battery());
 
         let tray = Main.__sm.tray;
         let elts = Main.__sm.elts;
diff --git a/system-monitor@paradoxxx.zero.gmail.com/metadata.json b/system-monitor@paradoxxx.zero.gmail.com/metadata.json
index 1851c46..37e481e 100644
--- a/system-monitor@paradoxxx.zero.gmail.com/metadata.json
+++ b/system-monitor@paradoxxx.zero.gmail.com/metadata.json
@@ -1,5 +1,5 @@
 {
-    "shell-version": ["3.26", "3.28", "3.30", "3.32", "3.34", "3.36", "40", "41", "42"],
+    "shell-version": ["3.26", "3.28", "3.30", "3.32", "3.34", "3.36", "40", "41", "42", "43"],
     "uuid": "system-monitor@paradoxxx.zero.gmail.com",
     "name": "system-monitor",
     "url": "https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet",
```

samueloph:
[...]

Issue #751, opened 2022-09-17
Compat with Ubuntu 22.10?


hadim:
It's marked as unsupported and when I force the installation, I don't see any monitoring informations in the topbar.

darkdragon-001:
Duplicate of #749 as Ubuntu 22.10 uses GNOME 43 by default.

glerroo:
This pull request #754 close this issue.

speedy-10:
I Just installed Ubuntu 22.10 on my laptop, but the extension doesn't work: 
Error: Tried to construct an object without a GType

How can I solve this?

System: 
  Kernel: 5.19.0-26-generic arch: x86_64 bits: 64 Desktop: GNOME v: 43.1
    Distro: Ubuntu 22.10 (Kinetic Kudu)
Graphics:
  Device-1: Intel TigerLake-LP GT2 [Iris Xe Graphics] driver: i915 v: kernel
  Device-2: Chicony Integrated Camera type: USB driver: uvcvideo
  Display: x11 server: X.Org v: 1.21.1.4 with: Xwayland v: 22.1.3 driver:
    X: loaded: modesetting unloaded: fbdev,vesa gpu: i915
    resolution: 1920x1200~60Hz
  OpenGL: renderer: Mesa Intel Xe Graphics (TGL GT2) v: 4.6 Mesa 22.2.1

glerroo:
This extension is not updated, you can install this that is an updated version [system-monitor-next](https://extensions.gnome.org/extension/3010/system-monitor-next/).

glerroo:
I use it on Xorg, have you restarted Gnome Shell (logout, loggin).
I think that on Wayland work because you logged out and than logged in a wayland session.

speedy-10:
your are right, login was helpful.

PR #754, opened 2022-10-19
Gnome-Shell 43 compatibility update.

glerroo:
Close issue: #749 and #751.

### DIFF
--- a/system-monitor@paradoxxx.zero.gmail.com/extension.js
+++ b/system-monitor@paradoxxx.zero.gmail.com/extension.js
@@ -1082,7 +1082,11 @@ const Battery = class SystemMonitor_Battery extends ElementBase {
         this.icon_hidden = false;
         this.percentage = 0;
         this.timeString = '-- ';
-        this._proxy = StatusArea.aggregateMenu._power._proxy;
+        if (shell_Version => '43') {
+            this._proxy = StatusArea.quickSettings._system._systemItem._powerToggle._proxy;
+        } else {
+            this._proxy = StatusArea.aggregateMenu._power._proxy;
+        }
         if (typeof (this._proxy) === 'undefined') {
             this._proxy = StatusArea.battery._proxy;
         }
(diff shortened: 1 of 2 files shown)


GaryElshaw:
Unfortunately i'm still seeing this on Ubuntu: 'Tried to construct an object without a GType'.

system-monitor@paradoxxx.zero.gmail.com/extension.js
```diff
@@ -1082,7 +1082,11 @@ const Battery = class SystemMonitor_Battery extends ElementBase {
         this.icon_hidden = false;
         this.percentage = 0;
         this.timeString = '-- ';
-        this._proxy = StatusArea.aggregateMenu._power._proxy;
+        if (shell_Version => '43') {
```
mgalgs, line 1085: This is actually just defining an arrow function, which will always evaluate to true. I believe you want `shell_Version >= '43'`)

mgalgs (changes requested):

mgalgs (commented):
lgtm. Has this been tested on 43 as well as something before 43?

system-monitor@paradoxxx.zero.gmail.com/extension.js
glerroo: Yes you are right, i maked a typo, i have just corrected it.
Thanks

mgalgs: :+1:

mgalgs:
For anyone who may be interested, I've integrated these patches into `system-monitor-next`:

https://github.com/mgalgs/gnome-shell-system-monitor-applet/pull/26#issuecomment-1304926321

averissimo:
Brilliant! I confirm that this is working on gnome 43.1 (Ubuntu 22.10 Kinetic)

jens-br:
Thanks for this! :+1: Since this PR was created two months ago: Is there a rough time frame when it will be integrated? This would be very helpful!

Many thanks again!

mgalgs:
@arderyp You can try https://extensions.gnome.org/extension/3010/system-monitor-next/, which is an extension built from [this fork](https://github.com/mgalgs/gnome-shell-system-monitor-applet) of this repo and contains this fix.

PR #755, opened 2022-10-23
Fix Travis tests

abitrolly:
Using newer Linux allows newer node which needs newer GLIBC

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 # https://docs.travis-ci.com/user/multi-os/
 # https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments
-dist: xenial
-sudo: required
+dist: focal
 language: node_js
 node_js: node
 


ZimbiX (approved):

chrisspen:
Looks like this might not have actually fixed the build. I tried merging it into PR 771, and it gets further, but it looks like it just times out.

ZimbiX:
Damn, I saw [`No errors detected`](https://app.travis-ci.com/github/paradoxxxzero/gnome-shell-system-monitor-applet/builds/256987351#L8010) and thought it was all good. It's doing the [same thing now](https://app.travis-ci.com/github/paradoxxxzero/gnome-shell-system-monitor-applet/builds/259907795#L8009) - is it actually timing out? The build does take over ten minutes :thinking:

abitrolly:
I did what I could. The next step is for somebody to try to build the project locally and report if there are errors there.

ZimbiX:
@chrisspen You had me there for a while thinking it was timing out; but I think it has indeed actually worked =P

Travis CI's logs retention seems to be a year - the oldest commit on master for which there's logs is f420845284d7c3083ce867f5d157d4e229f152d5 from 2022-02-04, and that build [_does_ time out](https://app.travis-ci.com/github/paradoxxxzero/gnome-shell-system-monitor-applet/builds/245908145#L8174):

```
Starting timeout countdown.
Stale for 2.861990451812744 seconds.
[...]
Stale for 117.97778844833374 seconds.
[...]
[  *** ] A stop job is running for TigerVNC …e desktop service (58s / 1min 30s)
Killing stale process...
```

The `Stale`/`Killing` logging comes from [`docker/run_and_kill.py`](https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/blob/ea55c44eabf470623b198ad5996a1635d2bbdcac/docker/run_and_kill.py#L52-L65). `Killing stale process...` means it timed out, whereas `Process has exited.` means it didn't time out (which is what we want).

In the latest two builds ([1](https://app.travis-ci.com/github/paradoxxxzero/gnome-shell-system-monitor-applet/builds/259907795#L7982), [2](https://app.travis-ci.com/github/paradoxxxzero/gnome-shell-system-monitor-applet/builds/259907815#L7983)), we see `Process has exited.`, which is good. And then the test output at the bottom indicates they ran successfully =)

It's unfortunate that the script does not cause the build to fail when it detects a timeout, which hid the failure of the old build.

I think what had us confused is that the output is somewhat misleading - we see logging that something is stale, and then we see Ubuntu shutting down before the test seems to run. It seems that some amount of stale messages are expected while the container starts up. And the test does actually run before Ubuntu shuts down; it's just that the output is stored in a file and only logged right at the end.

The structure of the Dockerised testing is rather complicated, and I had trouble keeping it all in my head, so had to sketch out a rough outline to keep track of it all. I've used host paths and paraphrased commands for simplicity.

- Travis CI - `.travis.yml`
    + `test.sh`
        * `docker/build-docker.sh`
            - `docker build -t gnome-shell-system-monitor-applet -f ./docker/Dockerfile.ubuntu2004 .`
        * `docker/run_and_kill.py`
            - Run with 120 second timeout:
                + `run-docker.sh`
                    * `docker run [...] gnome-shell-system-monitor-applet`, which must somehow run:
                        - `docker/ubuntu/xstartup`
                            + Backgrounds: `nohup testing/run_gui_tests > testing/stdout.log 2> testing/stderr.log &`
                            + `dbus-launch --exit-with-session gnome-session`
                            + Somehow it must detect when the tests finish, and close
        * `docker/check_output.sh`
            - Outputs `testing/stdout.log`
            - Checks that `testing/stderr.log` is empty; otherwise outputs it and `exit 1`

I'm having trouble getting the tests running locally, which is ironic for a Dockerised test system:

```
➜ ./docker/run-docker.sh
systemd 245.4-4ubuntu3.19 running in system mode. (+PAM +AUDIT +SELINUX +IMA +APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +SECCOMP +BLKID +ELFUTILS +KMOD +IDN2 -IDN +PCRE2 default-hierarchy=hybrid)
Detected virtualization docker.
Detected architecture x86-64.

Welcome to Ubuntu 20.04.5 LTS!

Set hostname to <c611a5c54894>.
Couldn't move remaining userspace processes, ignoring: Input/output error
Failed to determine supported controllers: No such file or directory
Failed to allocate manager object: No such file or directory
[!!!!!!] Failed to allocate manager object.
Exiting PID 1...
```

I'm guessing some assumptions are made about the host system, which my Arch system doesn't meet? I was surprised to see that the `docker run` command has so many options. I've used Docker quite a bit, but I'm not familiar with most of what this does.

https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/blob/ea55c44eabf470623b198ad5996a1635d2bbdcac/docker/run-docker.sh#L10-L16

Running `testing/test_gui.py` on my host doesn't work either.

```
➜ testing/test_gui.py
Traceback (most recent call last):
  File "/home/brendan/Projects/000_Others/gnome-shell-system-monitor-applet-paradoxxxzero/testing/test_gui.py", line 9, in <module>
    from dogtail import logging, predicate, rawinput, tree, utils
ModuleNotFoundError: No module named 'dogtail'
```

I don't see a Python packages lockfile, only:

https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/blob/ea55c44eabf470623b198ad5996a1635d2bbdcac/docker/Dockerfile.ubuntu2004#L117

I could probably get that working, but I'd prefer to have a Dockerised version given that should run the same on CI.

In doing some searching for how others have their CI set up for testing a Gnome Shell extension, I came across [this](https://www.reddit.com/r/gnome/comments/t5lxkm/enable_continuous_integration_tests_for_your/). The [Part III blog post](http://schneegans.github.io/tutorials/2022/03/02/gnome-shell-extensions-ci-03) linked there details a reasonably elegant setup. He prebuilds images for the different versions of Fedora (and corresponding versions of Gnome Shell). Switching to using those would take over the heavy lifting of getting Gnome running in Docker and pretty much just let us focus on the tests themselves, which would be nice.

abitrolly:
> I think what had us confused is that the output is somewhat misleading

I bet that's a standard problem that Python on Linux buffers output by default. https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/pull/776 should make printed messages appear immediately.

abitrolly:
> I'm having trouble getting the tests running locally, which is ironic for a Dockerised test system:

It uses some unsafe flags, indeed. Looks like https://github.com/darkdragon-001/Dockerfile-Ubuntu-Gnome/issues/2 is the best place to ask why they are needed.

chrisspen:
> You had me there for a while thinking it was timing out; but I think it has indeed actually worked =P

Yeah, looks like I spoke too soon. I'll go through and update the outstanding PRs.

chrisspen:
> I'm guessing some assumptions are made about the host system, which my Arch system doesn't meet? I was surprised to see that the `docker run` command has so many options. I've used Docker quite a bit, but I'm not familiar with most of what this does.

Yeah, unfortunately Docker isn't quite as host agnostic as most people think. Most of those options I don't fully understand either. They were recommended in various docs and blogs to work around various quirks limitations. I work on Ubuntu, so that's what it's optimized for.

>Running testing/test_gui.py on my host doesn't work either.

That assumes it's run from Docker, which installs Dogtail via pip, which I assume you haven't done running outside of Docker on your localhost.

If you setup a virtualenv and run the commands in the Dockerfile's "Set up GUI testing environment" section, that script should work for you, but I haven't tested it on Arch.

PR #775, opened 2023-01-21
Zimbi x add setting for tooltip delay

chrisspen:


ZimbiX:
I've rebased https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/pull/771, so this can be closed now

PR #776, opened 2023-01-21
Disable Python output buffering

abitrolly:
This should make build logs less confusing

https://stackoverflow.com/questions/107705/disable-output-buffering

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 set -e
 ./docker/build-docker.sh
-./docker/run_and_kill.py || true
+python -u ./docker/run_and_kill.py || true
 ./docker/check_output.sh


ZimbiX:
It doesn't look like this really changed anything. As I tried to describe in https://github.com/paradoxxxzero/gnome-shell-system-monitor-applet/pull/755#issuecomment-1399296855, the test log is only output in `docker/check_output.sh`, at the end of the build; it's not a line buffering issue.

I'm surprised this is so much more inelegant than in Ruby, where I usually set a similar thing within the code with `STDOUT.sync = true`

abitrolly:
@ZimbiX I am surprised that Linux people turn on buffering by default. If people want pipe performance, they could optimize for it explicitly. The default buffering just makes debugging hard.

abitrolly:
The best place to explain CI setup for testing GNOME extensions is GNOME docs, but this is all I could find https://blogs.gnome.org/shell-dev/2022/12/02/automated-testing-of-gnome-shell/

chrisspen:
> If people want pipe performance, they could optimize for it explicitly. The default buffering just makes debugging hard.

True, but the common default in most systems is to disable debugging, because unless things are horribly broken, it just adds a lot of extra wasted cycles. Buffering greatly speeds up scripts on systems with slow IO.

PR #780, opened 2023-01-24
Schema rebuild

mgalgs:
If people are checking the repo out from git and installing manually they
won't be getting the latest settings since the binary gschema file is
currently out of date.

This PR also adds a note about rebuilding the schema to the README.


mgalgs:
Cc: @ZimbiX

ZimbiX:
I was wondering what to do about this. Given it's generated with `make install` as per the installation instructions in the readme, I was thinking maybe it wasn't supposed to have been committed. I'd tried to leave it out of PRs to avoid merge conflicts.

mgalgs:
The readme has a section for manual installation for development (symlink rather than `make install`):

image

I agree that it's a bit unorthodox to have binary files committed. If we don't want to keep it updated in the repo we should delete it and add a `make extension` step to the section above. I'm good with either solution but as it stands the dev instructions don't actually result in a working dev environment.

ZimbiX:
Ah, I forgot that was there 😅 I've just been doing `VERSION=9999 make install` (setting the version to prevent auto-upgrade to the published version).

I'm not sure what's the norm. I guess until we decide otherwise, let's continue with how it is already? In which case, these added instructions are good.

If/when we do go with removing the binary files, it could be good to have something perform the generation of them automatically on change - using something like [guard](https://github.com/guard/guard).

ZimbiX (approved):

PR #792, opened 2023-05-18
Update README.md

Omcsesz:
Added instructions on how to install on CentOS7


ZimbiX (changes requested):
This addition isn't equivalent to the other commands in this section, as they're regarding installing the necessary dependencies (note this applet and the application called gnome-system-monitor are different things).

If you want to add this, the 'Repository installation' section is the place for it.

PR #802, opened 2025-09-27
Arch build test url

mgalgs:
AUR has been updated so we can point at that instead of my PKGBUILD fork. Opening this PR to ensure tests still pass in CI (already verified local).


mgalgs:
WRONG REPO! Holy cow :sweat_smile:

